PR #6, opened 2018-06-17
Add --update-all option

Enderlein:
Add option (--update-all, -u) to allow user to update all episodes of anime in their list at once. In the future, this should be changed to only update animes that are currently airing.


Enderlein:
I requested that the files be put in the wrong folder. Closed this to open another pull request.

PR #7, opened 2018-06-17
Add --update-all option that updates all anime on user's list

Enderlein:
This adds an option to "anime watch" (`--update-all, -u`) that updates all the anime on the user's list at once. It was just kinda annoying to update them one by one. In the future, this should only update anime that is currently airing so that you don't have to wait for every single thing on your list to update. My list is relatively small so it isn't a problem, but I can see the wait being annoying for someone with a larger list.

In order to do this I added a method called "anime_list" to the Watcher class, which returns the names of all the anime in the user's list, just to make things look a bit nicer in `cli.py`. However, it does feel kind of redundant. 

This also fixes a minor spelling mistake. "Specify" was spelled as "Specifiy" in line 46 of `cli.py`

### DIFF
--- a/anime_downloader/watch.py
+++ b/anime_downloader/watch.py
@@ -1,4 +1,4 @@
-from anime_downloader import config
+﻿from anime_downloader import config
 from anime_downloader.sites.nineanime import NineAnime
 
 import os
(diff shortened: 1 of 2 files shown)


anime_downloader/watch.py
```diff
@@ -43,6 +43,16 @@ def list(self):
...
 
+    def anime_list(self):
+        # Stores list of anime names in watcher's list
+        anime_names = []
+        animes = self._read_from_watch_file()
+
+        for idx, anime_name in enumerate(animes):
+            anime_names.append(anime_name)
+        
+        return anime_names
```
vn-ki: Won't `return self._read_from_watch_file()` be enough?

Enderlein: You're right. Fixed.

vn-ki:
Thank you for this. I'll merge it once the ci pass.

If you want take a dig at updating only airing anime, check out #8.

vn-ki:
The fail is due to 9anime changing something. I'll still merge this. :tada:

PR #9, opened 2018-06-17
Only update anime if it's status is "airing", Show number of retry attempts

Enderlein:
`update_anime` will only update anime on the list that is marked as "airing".
Also changed log warning under `watch_anime` function in `cli.py` to display which try they're currently on


vn-ki:
I'll merge this after fixing 9anime. 🙂

vn-ki:
@Enderlein Looks like 9anime changed their api yet again. :crying_cat_face:

PR #22, opened 2018-06-20
Fallback for 9anime using 9anime.cloud

vn-ki:
This Just Works.

I don't know if 9anime.cloud will stay working or not, but it looks like a reliable fallback for now.


vn-ki:
I'll merge this (because it's mostly safe) and make a new release ASAP.

PR #23, opened 2018-06-21
Filter out specials and other and fix watch

vn-ki:
Kissanime has recaps, specials, opening, ending, etc. For now, let's just filter them out.

Also `Watcher.update` would error in master because `Kissanime` has no `meta`. This fixes that too.


Enderlein:
Looks good. Merging.

PR #66, opened 2018-08-19
Added AnimePahe support

ngomile:
AnimePahe support has been successfully added though that is a misleading statement as the mp4upload extractor has not been implemented. I'm making this pull request to notify you of both the support of animepahe and lack of support for mp4upload extractor due to not knowing how to go about it. Thank you.


vn-ki:
As for mp4upload, if my memory serves correct, it uses js obfuscation. The easiest method would be execute the javascript in node js to get the stream url. But the best method would be to reverse the obfuscation so that there is no dependence on node.

I'll take a look into mp4upload later in the week. If reversing proves to be too hard, we'll take the node js way.

ngomile:
mp4upload isn't the only source for animepahe but it is the oldest one and still used. Basically if mp4upload is supported it means that we can go back to some of the oldest anime on animepahe and still manage to download them through mp4upload since that's what was being used then. Other sources include openload, streamango and kwik.

ngomile:
In relation to extracting from mp4upload I was even thinking of adding Selenium into the mix but I thought it would be a bit overkill. Though it would make extracting the link a breeze since if you're using a browser mp4upload brings up the video.

vn-ki:
@ngomile I understand you. Looks like mp4upload is the only way.

I would merge this after implementing mp4upload extractor only. I don't want an `Anime` class which does nothing. I've been a bit busy lately, but I think I can come back to it in a couple of weeks or so.

As for the usage of Selenium, I'm kind of against it. Selenium is huge and slow. I really don't want to go that way if I can. As I said, I'll take a look and revert back. 

If you find a way to extract the stream link before me, just update this PR.

ngomile:
Thanks for the response. I'll try and see what I can do but trying to figure out what mp4upload is doing is no easy task for me. But I'll try nonetheless and I didn't use Selenium for the exact reason you stated though it was very tempting.

vn-ki:
@ngomile Can you implement an extractor for kwik and support kwik for animepahe? Kwik seems to relatively easy.

That  way, I'll merge this in and we can work on mp4upload later.

ngomile:
I'll see what I can do but these guys do not make this any easier. There seems to be a lot of hiding but I'll have a go at kwik.

ngomile:
[Managed to find a project that implements an mp4upload extractor](https://github.com/py7hon/mp4upload-direct/blob/master/get.php). Using this I'll be able to implement the mp4upload extractor and credit will be given to my source of insight.

vn-ki:
Can you add animepahe to readme? 

Once that's done, I'll merge this.

ngomile:
Done.

vn-ki:
Thanks @ngomile. This is a very nice contribution.

PR #85, opened 2018-10-03
Fix downloading for masterani

freddiaN:
This should fix downloading from masterani/mp4upload for some shows (since it also seems that some anime, like Koe No Katachi, are not affected by broken mp4upload SSL certs).


anime_downloader/sites/anime.py
```diff
@@ -102,6 +102,7 @@ class BaseEpisode:
     QUALITIES = None
     title = ''
     stream_url = ''
+    ssl = True
```
vn-ki: You don't seem to be using these. Remove them.

anime_downloader/sites/masterani.py
```diff
@@ -12,6 +12,7 @@
 
 class MasteraniEpisode(BaseEpisode):
     QUALITIES = ['360p', '480p', '720p', '1080p']
+    ssl = False
```
vn-ki: Same as above. Remove since you are not using them.

anime_downloader/util.py
```diff
@@ -75,7 +75,6 @@ def search(query, provider):
 
     return url
 
-
```
vn-ki: Don't remove spaces. :)

anime_downloader/sites/anime.py
```diff
@@ -166,7 +167,7 @@ def _get_sources(self):
         raise NotImplementedError
 
     def download(self, force=False, path=None,
-                 format='{anime_title}_{ep_no}', range_size=None):
+                 format='{anime_title}_{ep_no}', range_size=None, ssl=ssl):
```
vn-ki: `ssl=True`, right?

anime_downloader/cli.py
```diff
@@ -140,7 +140,14 @@ def dl(ctx, anime_url, episode_range, url, player, skip_download, quality,
             if chunk_size is not None:
                 chunk_size *= 1e6
                 chunk_size = int(chunk_size)
-            episode.download(force=force_download,
+            if cls == Masterani:
```
vn-ki: Do 
```
cls.__name__ == 'Masterani'
```

anime_downloader/cli.py
```diff
@@ -4,6 +4,7 @@
 
 import logging
 
+from anime_downloader.sites.masterani import Masterani
```
vn-ki: Remove this import as you won't need it anymore after the change mentioned below.

freddiaN:
updated my fork to fix those issues (and also tested again, still worked)

PR #96, opened 2018-10-17
Multi fixes, some reversions and new functionality

ngomile:
Made Gogoanime functional, added a new way to collect episodes. Had to revert back to old code for animepahe and overall minor tweaks here and there.


anime_downloader/cli.py
```diff
@@ -30,6 +30,10 @@ def cli():
 @click.option(
     '--episodes', '-e', 'episode_range', metavar='<int>:<int>',
     help="Range of anime you want to download in the form <start>:<end>")
+@click.option(
```
vn-ki: I think we don't need a separate CLI argument for this. This functionality should be added to `--episodes-range` (It is not ranges anymore then, but I think we can ignore it). 

`util.split_anime` can be changed to facilitate this functionality. Maybe it's name should be changed to `parse_ep_str` (or something).

anime_downloader/sites/masterani.py
```diff
@@ -41,14 +41,12 @@ def _get_sources(self):
 
             ret.append((source['host']['name'].lower(), url))
 
-        # TODO: Clean this up. Why it looks like this, I don't know.
-        ret_1 = [(name, url) for name, url in ret if name in ['stream.moe']]
-        ret_1 += [(name, url) for name, url in ret if name == 'rapidvideo']
-        ret_1 += [(name, url) for name, url in ret if name == 'mp4upload']
+        sources = ['stream.moe', 'rapidvideo', 'mp4upload']
+        ret = [(name, url) for name, url in ret if name in sources]
```
vn-ki: Did you test masterani after this change? If not, please do.

ngomile: Yes I did, works exactly the same way.

anime_downloader/cli.py
ngomile: Indeed, I can change the name of the parse_grammar function to parse_episode_string. I would personally like to keep split_anime the way it is and have the new episode parser use it like it is now. I'll drop the option from the command line and place the new episode parser in.

ngomile:
Done some changes.

Issue #132, opened 2019-01-26
very specific(?) mp4upload error


rehtlaw:
**Describe the bug**

the extractor doesn't seem to be able to find the actual video file on the embedded mp4upload site [even tho it's definitely there](https://fred.moe/Eoy.png) and it has `video` in the name (which seems to be how the link is found)

**To reproduce**

```
anime dl 'https://www.masterani.me/anime/info/1792-rezero-kara-hajimeru-isekai-seikatsu' --disable-ssl -e 9:
```
I should also note that, at least for me, this only happened from the 9th episode on, all other 8 episodes downloaded just fine

<details>
<summary>LOG</summary>

cannot copy more of the logs, but every attempt at downloading an episode results in the same
```

DEBUG root: Domain: s1, Video ID: qsx7723gz3b4quuoyouugpchkiep7xfa5leb2zduxhz4y4pr3okpmw4a, Protocol: 282
DEBUG root: Title is ReZero_kara_Hajimeru_Isekai_Seikatsu_-_17_-_1080p
DEBUG root: Stream URL: https://s1.mp4upload.com:282/d/qsx7723gz3b4quuoyouugpchkiep7xfa5leb2zduxhz4y4pr3okpmw4a/ReZero_kara_Hajimeru_Isekai_Seikatsu_-_17_-_1080p.mp4
DEBUG root: Extracting stream info of id: https://www.masterani.me/anime/watch/1792-rezero-kara-hajimeru-isekai-seikatsu/18
DEBUG root: [cfscrape] Calling https://www.masterani.me/anime/watch/1792-rezero-kara-hajimeru-isekai-seikatsu/18
[...]

PR #134, opened 2019-01-28
Handle error when extracting MP4Upload stream URL

ngomile:
This fixes an issue where the MP4Upload extractor was failing hard when attempting to extract its stream url such as issue #132.


ngomile:
Alright I'll close this then.

PR #161, opened 2019-03-22
Rewrite

vn-ki:
This is the rewrite. 

Should be functional. Some more stuff to do.

Everyone is invited to test out and give feedback.


ngomile:
This can be something that could possibly be improved with this rewrite, currently the BaseAnime class and the BaseEpisode class know too much about each other. It would be nice if they were properly separated not to be so reliant on one another. For example instead of BaseEpisode having to find out what its title is from BaseAnime it can receive the title as an argument to its constructor.

Such things as deciding what to do with the given fallback qualities can also be passed down to the BaseEpisode class, I've had some difficulty trying to refactor some things since they are so glued together. Again, just something to consider but things work so yeah.

vn-ki:
@ngomile  Thanks for the input.

> For example instead of BaseEpisode having to find out what its title is from BaseAnime it can receive the title as an argument to its constructor.

I would be against this. This would mean that every time an attribute has to be passed down, the constructor and the constructor call would have to be modified. (Passing down the parent is an actual design pattern, don't remember the name but I've seen it used in the flutter framework)

> given fallback qualities can also be passed down

Following the above comment, I think the fallback qualities and qualities should only be in `Anime` class and not `Episode` class. The episode class when it needs can do `Episode._parent.anything`.

**NOTE TO SELF**: Remove fallback qualities and qualities from episode class.

vn-ki:
**NOTE TO SELF**: Figure out a better way to handle qualities not found and fallback qualities.

vn-ki:
@ngomile More comments like this are welcome.

ngomile:
@vn-ki no problem at all. I'll try to take a harder look at more areas of the code base that may need a rewrite. And on the Flutter pattern, I think you might be talking about how the context of the parent widgets (I have no idea if I am correct on what I am saying so please bare with it) can be passed down to the children.

This works for Flutter and of course is working right here but maybe there's a simpler way though more suited to Python. But nonetheless of course, if it ain't broke don't fix it. And on another note though this can be done right now I think it would be nice to also put the available sites in a centralized place.

So instead of having to go to the two areas that need the sites A.K.A the watch and download options we can just cut the sites out of that list of options, put it in a variable and that is all. I'll comment on the place I am talking about but you probably know it. Again I'll look for more things but maybe the rewrite will be done by then.

anime_downloader/commands/dl.py
```diff
@@ -0,0 +1,126 @@
...
+    help='Force downloads even if file exists')
+@click.option(
+    '--file-format', '-ff', default='{anime_title}/{anime_title}_{ep_no}',
+    help='Format for how the files to be downloaded be named.',
+    metavar='FORMAT STRING'
+)
+@click.option(
+    '--provider',
+    help='The anime provider (website) for search.',
+    type=click.Choice(['9anime', 'kissanime', 'twist.moe', 'animepahe', 'kisscartoon', 'masterani', 'gogoanime'])
```
ngomile: @vn-ki  this is the part I meant, it would be easier if it was just put into its own variable. Simple change but convenient.

ngomile:
@vn-ki  Just a heads up but it seems that [cfscrape is no longer supported ](https://github.com/Anorov/cloudflare-scrape/issues/231). So due to the issues that have been arising from failed cloudflare bypassing the now supported cloudflare scraper may yield better results.

29ayush:
> @vn-ki Just a heads up but it seems that [cfscrape is no longer supported ](https://github.com/Anorov/cloudflare-scrape/issues/231). So due to the issues that have been arising from failed cloudflare bypassing the now supported cloudflare scraper may yield better results.

I wasn't able to access kissanime using `cfscrape` but `cloudscraper` worked. It's the one ngomile has linked.

vn-ki:
I'm just wondering whether I should fix v3 or not. Fixing cfscrape issue in rewrite is easy as pie. But in master, it's a different story.

Thanks for the heads up anyway.

vn-ki:
Looks like they are fixing cfscrape. I'll just wait it out then.

Issue #154, opened 2019-03-11
Kissanime search doesn't work at all


Arnav71:
![wtf](https://user-images.githubusercontent.com/16363526/54123694-6080a380-4426-11e9-84a2-b2b44cb90b5e.png)

(See the image)

In this example I tried searching for Naruto,but it returned an anime not related to it.

meyerovb:
did you try it without add the provider param?  just anime dl 'Naruto'

vn-ki:
Is this still hapenning? If yes use the `rewrite` branch and post the output of the log.

Arnav71:
Yup,(sorry but how do I use the rewrite branch??)

meyerovb:
@Arnav71 in github go to the code tab and change the branch dropdown.  you can download that code and run the program from the new folder

meyerovb:
@vn-ki i did this:

pip uninstall anime-downloader

pip install git+https://github.com/vn-ki/anime-downloader.git@rewrite#egg=anime-downloader[cloudflare]

anime dl --provider kisscartoon -xd "{aria2}" hotel

and got this:

ERROR: No module named 'anime_downloader.commands'

Guess I'm not sure how to run the rewrite branch either

vn-ki:
This is probably fixed in master. If not please reopen

Issue #128, opened 2019-01-16
Kissanime Not Working


as280093:
today i tried the anime dl command with url but it returns

 
![screenshot](https://user-images.githubusercontent.com/25240839/51232937-7d50ab80-198e-11e9-9b17-f929165e0a16.png)

it ends up like this for the first time i have to try 9anime looks like for time being 9anime should work but can anyone see whats the issue

IngwiePhoenix:
```
Ingwie@Ingwies-Macbook-Pro.local /tmp $ anime dl "https://kissanime.cool/watch/akame-ga-kill-dub.7x07/y0zp51"
anime: anime-downloader 3.6.2
anime: Extracting episode info from page
'NoneType' object has no attribute 'find'
Ingwie@Ingwies-Macbook-Pro.local /tmp $ pip3 --version
pip 18.1 from /usr/local/lib/python3.7/site-packages/pip (python 3.7)
Ingwie@Ingwies-Macbook-Pro.local /tmp $ python --version
Python 2.7.10
Ingwie@Ingwies-Macbook-Pro.local /tmp $ python3 --version
Python 3.7.2
```

That doesn't seem right either. Any ideas? I installed `anime-downloader` through `pip3`.

vn-ki:
Is this still happening?

as280093:
> Is this still happening?

yess

vn-ki:
Need more info. Please post logs.

kmolivier:
I have the same issue

vn-ki:
Note to self: Kissanime returns `Not Found`. Investigate why exception is not reaised

vn-ki:
Can you check whether this happens on master? @as280093 @IngwiePhoenix

IngwiePhoenix:
I previously used `pipx` to install this. so how exactly do I run it off Master? I am not a snake guy... ^^;

vn-ki:
This should be fixed in master.

@IngwiePhoenix you can hop on to discord for help with installing the master version.

PR #169, opened 2019-04-01
Fix KissAnime search,download and exception handling.

29ayush:
#154 When I changed the protocol to https it worked fine.
I can see that in rewrite you have already used it as https so this problem shouldn't be there (but I haven't checked.).

#128 Added as file comments.

Do change error name as per your liking.


nandahkrishna:
The KissAnime issue still persists. Can you add tests to avoid the coverage fail @29ayush?

vn-ki:
Is this still required @29ayush ?

vn-ki:
I think this has been superseded by the rewrite. Therefore, closing.

Issue #198, opened 2019-05-30
KeyError: 'playlist'


jcurrenc:
KeyError: 'playlist'

I was able to download the first 14 episodes with no issue, no episodes past this will download.

**To reproduce**

```
anime dl https://kisscartoon.is/Cartoon/Mia-and-Me-Season-3 -e 15:26 -ll DEBUG -xd "{aria2}"
```

<details>
<summary>LOG</summary>

```
PS C:\MiaS3> anime dl https://kisscartoon.is/Cartoon/Mia-and-Me-Season-3 -e 15:26 -ll DEBUG -xd "{aria2}"
INFO root: anime-downloader 3.6.3
DEBUG root: Platform: Windows-10-10.0.17134-SP0
DEBUG root: Python 3.7.3
INFO root: Extracting episode info from page
[...]

PR #239, opened 2019-10-04
Updated episode list url

cmjesz:
Fixes vn-ki/anime-downloader#198


vn-ki:
I don't think it's that easy. The new API seems to have different behavior. Did you test this change?

PR #245, opened 2019-10-18
Fix streaming from twist.moe

LiteracyFanatic:
Streaming from twist.moe is currently broken for me. Changing https://eu1.twist.moe to https://www.twist.moe, deleting the shows from `watch.json` and adding them again with `anime watch --new`, and passing `--http-header-fields="referer: https://twist.moe/"` to `mpv` fixes the issue. I also made `BasePlayer` accept an `AnimeEpisode` instead of a URL so that the `sitename` can be used to determine whether or not the extra argument to `mpv` is needed.

### DIFF
--- a/anime_downloader/players/mpv.py
+++ b/anime_downloader/players/mpv.py
@@ -19,7 +19,9 @@ class mpv(BasePlayer):
 
     @property
     def args(self):
-        return ['--input-conf='+get_mpv_configfile(), self.stream_url]
+        return ['--input-conf='+get_mpv_configfile(),
+                '--http-header-fields=referer: '+self.episode.source().referer,
+                self.episode.source().stream_url]
 
 
 def get_mpv_home():
--- a/anime_downloader/commands/watch.py
+++ b/anime_downloader/commands/watch.py
@@ -200,7 +200,7 @@ def watch_anime(watcher, anime):
                 'Playing episode {}'.format(episode.ep_no)
             )
             try:
-                player = mpv(episode.source().stream_url)
+                player = mpv(episode)
             except Exception as e:
                 anime.episodes_done -= 1
                 watcher.update(anime)
--- a/anime_downloader/players/baseplayer.py
+++ b/anime_downloader/players/baseplayer.py
@@ -31,10 +31,10 @@ class BasePlayer(metaclass=ABCMeta):
         '''
         return
 
-    def __init__(self, stream_url):
+    def __init__(self, episode):
         # TODO: Stream urls is a list of urls for now
         # It should be a list of seperate class with title and other metadata
-        self.stream_url = stream_url
+        self.episdoe = episode
 
     def _get_executable(self):
         if os.name == 'nt':
--- a/tests/test_player.py
+++ b/tests/test_player.py
@@ -2,19 +2,19 @@ from anime_downloader.players.mpv import mpv, get_mpv_configfile
 import os
 
 
-class TestMPV:
-    def setup(self):
-        self.player = mpv('example.mp4')
+# class TestMPV:
+#     def setup(self):
+#         self.player = mpv('example.mp4')
 
-    def test_args(self):
-        assert len(self.player.args) == 2
+#     def test_args(self):
+#         assert len(self.player.args) == 2
 
-    def test_exe(self):
-        exe = self.player._get_executable()
-        if os.name == 'nt':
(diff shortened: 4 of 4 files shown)


LiteracyFanatic:
It's worth noting that this pull request is a superset of #156.

vn-ki:
I have some ideas on how to handle referrers better. This feels like a hack and I don't want to accept it.

If you have some opinions on this or have some suggestion feel free to share.

LiteracyFanatic:
Perhaps a new class called something like `StreamInfo` could be created which would contain the URL and a dictionary of HTTP headers and values. This would replace the `stream_url` field of `AnimeEpisode`. `TwistMoeEpisode` would add the referer header and any of the other sites' episode classes could add any necessary headers of their own instead. Then any classes inheriting from `BasePlayer` would just have to implement the proper command line argument formatting for the headers .

anime_downloader/players/mpv.py
```diff
@@ -19,7 +19,12 @@ def _get_executable_posix(self):
 
     @property
     def args(self):
-        return ['--input-conf='+get_mpv_configfile(), self.stream_url]
+        _args = []
+        _args.append('--input-conf='+get_mpv_configfile())
+        if self.episode.sitename == "twist.moe":
```
vn-ki: Can you change this to use `self.episode.source().referer`?

With the recent update I made, this should work(?).

LiteracyFanatic: I can confirm that it does work with this change made. Passing an episode instead of a URL breaks one of the tests for mpv though. Fixing it would either involve creating an actual anime and episode with one of the sites or implementing some sort of mock objects. The first method would be more tightly coupled than is ideal for a unit test. I took a stab at modifying the base classes to allow them to be instantiated without an actual site but was unsuccessful.

anime_downloader/extractors/mp4upload.py
```diff
@@ -18,22 +18,25 @@ def _get_data(self):
...
+        logger.debug('Domain: %s, Video ID: %s, Protocol: %s' % (domain, video_id, protocol))
 
         url = self.url.replace('embed-', '')
         # Return to non-embed page to collect title
-        mp4u_page = helpers.soupify(helpers.get(url).text)
-
-        title = mp4u_page.find('span', {'class': 'dfilename'}).text
-        title = title[:title.rfind('_')][:title.rfind('.')].replace(' ', '_')
+        mp4u_page = helpers.get(url, referer=self.url).text
+        title = title_re.search(mp4u_page)
```
vn-ki: What are these? Looks like you messed up git history a bit here.

anime_downloader/players/mpv.py
vn-ki: Comment out the test for now and open an issue. We can take care of that later.

LiteracyFanatic: Will do.

LiteracyFanatic: Created issue #303.

anime_downloader/extractors/mp4upload.py
LiteracyFanatic: Woops, no idea how I did that. I'll rebase onto master and clean that up.

LiteracyFanatic:
After rebasing onto master this no longer works since the watch command is deprecated. `anime dl --player mpv` doesn't work either, because it just uses the argument as the name of an executable and passes the URL to it. I can fix that by adding logic to figure out what subclass of `BasePlayer` to instantiate based on the argument. If adl is the recommended solution for streaming, is this pull request even still relevant? It might make more sense to just make sure adl is passing the right headers to mpv when it encounters a twist URL. Thoughts?

vn-ki:
I do want anime watch functionality retained. Even though I have deprecated it, I have not removed it.

`anime watch` was written almost 2 years ago and was never improved. It is a huge mess and the only way to fix it is a full rewrite.

I think `anime dl --player mpv` should be made to use Player class.

> If adl is the recommended solution for streaming, is this pull request even still relevant?

Yes. Down the line, I want anime watch to be revived.

> It might make more sense to just make sure adl is passing the right headers to mpv when it encounters a twist URL.

That makes sense too, at least for now. Pinging @RaitaroH

vn-ki:
Can you rebase this?

LiteracyFanatic:
Ok, that all makes sense. I've rebased onto master.

Issue #80, opened 2018-09-21
Output URLs to file


Meisterlala:
It would be nice if the stream URL would be writen to a file.

Currently when you type `anime dl "overlord" --url > url.txt` the url.txt contains a list of the animes you can select and you have to blindly choose one. Text which is prefixed with "anime: " is still displayed in the console.

You could add another option, e.g. `anime dl "overlord" --url "url.txt"` which writes the stream URLs to that text file. Or you could use the pipe system, e.g. `anime dl "overloard" --url > url.txt` and only output the stream URLs to STDOUT.

vn-ki:
This seems to be a bug. `click.echo` seems to be outputting to stdout instead of stderr.

PR #250, opened 2019-10-18
Fix on anime dl --url flag writing only anime url when provided output file redirection

nulligor:
RELATED ISSUE: https://github.com/vn-ki/anime-downloader/issues/80

Desired output and behavior can be achieved when the command is evoked as: 
`anime -ll ERROR dl "overlord" --url > url.txt`

### DIFF
--- a/anime_downloader/util.py
+++ b/anime_downloader/util.py
@@ -77,16 +77,16 @@ def search(query, provider, choice=None):
     # cli. But it is used in watch too. :(
     cls = get_anime_class(provider)
     search_results = cls.search(query)
-    click.echo(format_search_results(search_results))
+    click.echo(format_search_results(search_results), err=True)
 
     if not search_results:
         logger.error('No such Anime found. Please ensure correct spelling.')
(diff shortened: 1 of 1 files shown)


anime_downloader/util.py
```diff
@@ -72,18 +72,18 @@ def format_search_results(search_results):
     return table
 
 
-def search(query, provider):
+def search(query, provider, stderr=False):
```
vn-ki: You don't have to add this extra parameter

anime_downloader/util.py
```diff
@@ -72,18 +72,18 @@ def format_search_results(search_results):
...
     # Since this function outputs to stdout this should ideally be in
     # cli. But it is used in watch too. :(
     cls = get_anime_class(provider)
     search_results = cls.search(query)
-    click.echo(format_search_results(search_results))
+    click.echo(format_search_results(search_results), err=stderr)
```
vn-ki: Just do `err=True`

PR #253, opened 2019-10-21
Gogoanime Default DL Host Fix

andrewlef:
Switched download host to gogo default hosting servers (cdnfile).  This should pre-empt the issues experienced with intermittent improper mp4upload url encoding and rapidvideo unavailability.  Additionally, it also fixes the problem of low quality selection.  Using the default hoster will always download at original (highest) quality.

As of now, mp4upload and rapidvideo are disabled (commented), but can easily be restored by uncommenting.  I also inserted code for streamango support, though it's currently disabled as well.

Ideally, there should be a way to give the user a choice of host server (cdnfile, mp4, rapid, etc) to download from.  Or perhaps a method of auto fallback when an error occurs.  The current error handling isn't ideal.  I haven't implemented this, as I'm unsure of the best way to do so.  Having an additional prompt for each file isn't ideal.


vn-ki:
Can you remove the additional .DSStore file?

andrewlef:
Sorry about that.  MacOS likes to insert those everywhere.  Looks like you removed it before I could though.

Issue #258, opened 2019-10-26
errorCode=19 Failed to resolve the hostname ended.mp4upload.com


andrewlef:
**Describe the bug**

The mp4upload helper is incorrectly scraping the download link.  Specifically, the 'domain' variable in 'source_parts_re' is resolving to the value "ended".  This is incorrect.  In the example below, the variable should resolve to "www14". 

**To reproduce**

```
anime -ll DEBUG dl "stone"  -q 720p  -e 1 --provider gogoanime
```

<details>
<summary>LOG</summary>

```
iMac:~ user$ anime -ll DEBUG dl "stone"  -q 720p  -e 1 --provider gogoanime
2019-10-26 10:49:34 iMac.local anime_downloader.util[98177] INFO anime-downloader 4.0.1
2019-10-26 10:49:35 iMac.local anime_downloader.util[98177] DEBUG Platform: Darwin-18.5.0-x86_64-i386-64bit
2019-10-26 10:49:35 iMac.local anime_downloader.util[98177] DEBUG Python 3.7.4
2019-10-26 10:49:35 iMac.local anime_downloader.sites.helpers.request[98177] DEBUG HTML file temp_dir: /var/folders/sy/f41bjb5d2v743qrsz51t_z680000gn/T/animedlou2bxhm2
2019-10-26 10:49:35 iMac.local anime_downloader.sites.helpers.request[98177] DEBUG -----
2019-10-26 10:49:35 iMac.local anime_downloader.sites.helpers.request[98177] DEBUG GET https://ajax.apimovie.xyz/site/loadAjaxSearch
2019-10-26 10:49:35 iMac.local anime_downloader.sites.helpers.request[98177] DEBUG {'params': {'keyword': 'stone', 'id': -1, 'link_web': 'https://www1.gogoanime.sh/'}}
2019-10-26 10:49:35 iMac.local anime_downloader.sites.helpers.request[98177] DEBUG {'user-agent': 'Mozilla/5.0 (Windows; U; Windows NT 5.1; en-US) AppleWebKit/532.1 (KHTML, like Gecko) Chrome/4.0.219.0 Safari/532.1'}
2019-10-26 10:49:35 iMac.local anime_downloader.sites.helpers.request[98177] DEBUG -----
[...]

Issue #252, opened 2019-10-20
NoneType object is not subscriptable


ESPGranEdu:
**Describe the bug**
Getting NoneType object is no subscriptable with a specific provider

**To reproduce**

```
anime dl "Steins;Gate 0" --provider animeflv
```

<details>
<summary>LOG</summary>

```

[edupy@archlinux-edupy → Steins;Gate 0]$ anime -ll DEBUG dl "Steins;Gate 0" --provider animeflv
2019-10-20 19:14:07 archlinux-edupy anime_downloader.util[2155] INFO anime-downloader 4.0.0
2019-10-20 19:14:07 archlinux-edupy anime_downloader.util[2155] DEBUG Platform: Linux-5.3.7-arch1-1-ARCH-x86_64-with-arch
2019-10-20 19:14:07 archlinux-edupy anime_downloader.util[2155] DEBUG Python 3.7.4
2019-10-20 19:14:07 archlinux-edupy anime_downloader.sites.helpers.request[2155] DEBUG HTML file temp_dir: /tmp/animedlohk_igd0
2019-10-20 19:14:07 archlinux-edupy anime_downloader.sites.helpers.request[2155] DEBUG -----
2019-10-20 19:14:07 archlinux-edupy anime_downloader.sites.helpers.request[2155] DEBUG GET https://animeflv.net/browse?q=Steins;Gate 0
2019-10-20 19:14:07 archlinux-edupy anime_downloader.sites.helpers.request[2155] DEBUG {}
2019-10-20 19:14:07 archlinux-edupy anime_downloader.sites.helpers.request[2155] DEBUG {'user-agent': 'Mozilla/5.0 (Macintosh; AMD Mac OS X 10_8_2) AppleWebKit/535.22 (KHTML, like Gecko) Chrome/18.6.872'}
2019-10-20 19:14:07 archlinux-edupy anime_downloader.sites.helpers.request[2155] DEBUG -----
send: b'GET /browse?q=Steins;Gate%200 HTTP/1.1\r\nHost: animeflv.net\r\nuser-agent: Mozilla/5.0 (Macintosh; AMD Mac OS X 10_8_2) AppleWebKit/535.22 (KHTML, like Gecko) Chrome/18.6.872\r\nAccept-Encoding: gzip, deflate\r\nAccept: */*\r\nConnection: keep-alive\r\n\r\n'
reply: 'HTTP/1.1 200 OK\r\n'
[...]

PR #259, opened 2019-10-26
Fixed Incorrect mp4upload Parsing & Animeflv downloading

andrewlef:
See Issue #258 and Issue #252.

### DIFF
--- a/anime_downloader/config.py
+++ b/anime_downloader/config.py
@@ -33,7 +33,7 @@ DEFAULT_CONFIG = {
         },
         'animeflv': {
             "version": "subbed",
-            "server": "natsuki",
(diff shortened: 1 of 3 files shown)


vn-ki:
Why did you remove natsuki?

andrewlef:
Because, as far as I can tell, animeflv no longer has a content host server called natsuki.  The only servers I saw were mega, zippyshare, openload and streamango.

andrewlef:
Ok, I made some additional changes to the code and added some error handling with a descriptive error message.  That should eliminate certain bug reports.  

Also noticed that the build fails the auto test.  Builds just fine for me in Python 3.7 on Mac, though.

PR #261, opened 2019-10-30
Fix typo in Pipfile

ajyong:
The `python-language-server` dev dependency had a typo.  Now that it's fixed, `pipenv install` should work again.

I believe the `Pipfile.lock` file should be included in this PR, but if it breaks your setup, I can remove it.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -14,4 +14,4 @@ radon = "*"
 "flake8" = "*"
 httpretty = "*"
 pylint = "*"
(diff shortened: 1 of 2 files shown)


chenrui333:
ping @vn-ki, seems a good fix.

ajyong:
Rebased my branch since it's been awhile.  Diff is the same though.

vn-ki:
Thanks. Spelling is hard! XD

Issue #263, opened 2019-11-01
Streamango is now dead


andrewlef:
**Describe the bug**

Streamango is dead.  Need to remove it from code to prevent errors.

**To reproduce**

```

```

<details>
<summary>LOG</summary>

```



```

MichaelDavidJohnson:
Looks like you did it

Issue #254, opened 2019-10-23
Rapidvideo is dead


andrewlef:
Now that Rapidvideo is gone, it probably makes sense to deprecate the associated rapidvideo code.

MichaelDavidJohnson:
Close this.

PR #264, opened 2019-11-01
Disable Streamango and Rapidvideo

andrewlef:
Went through and disabled the code for streamango and rapidvideo.  Both sites are dead.  Also disabled Kissanime, as the captcha has broken the site for dl.

See issues: #254 and #263 

I also fixed gogoanime so it handles quality choice (and fallback qualities) for the default host (cdn).


MichaelDavidJohnson:
Close, this is from 8 months ago now.

PR #265, opened 2019-11-02
Specify to download sub only

mcj64:
downloader may obtain link from faststream provider which the link does not work or obtain link that downloads dubbed version of the anime


andrewlef:
Forcing the app to download subs only is not a good design.  Can you implement a design that would enable to user to choose between sub and dub?

mcj64:
> Forcing the app to download subs only is not a good design. Can you implement a design that would enable to user to choose between sub and dub?

You would need to ask vn-ki or the person who added animeflix since that would need to change many things which i am cant do as a beginner in programming. One thing though is that animeflix mix both sub and dub version of the anime in a same api link so you gonna have either ones when requesting from the api.

MichaelDavidJohnson:
Can this be closed due to inactivity.

Blatzar:
Closed due to inactivity

Blatzar:
Close

MichaelDavidJohnson:
Close

PR #298, opened 2020-03-20
Dubbedanime

vn-ki:
@Blatzar

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,9 +21,9 @@
 </a>
 </div>
 
-  
+
 ---
-  
+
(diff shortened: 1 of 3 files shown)


Blatzar:
One regret when reading the code, instead of split, I could've just used .replace('"trollvid","id":"',''), that's probably more readable
```
            token = re.search(tokenregex,soup).group(0).split('"trollvid","id":"')[1]
            url = re.search(r'source src="[^"]*',soup).group().split('source src="')[1]
```

PR #305, opened 2020-03-22
use rapidfuzz instead of fuzzywuzzy

maxbachmann:
FuzzyWuzzy is GPLv2 licensed which would force you to licence the whole project under GPLv2. I had the same problem on one of my projects and so I wrote [rapidfuzz](https://github.com/rhasspy/rapidfuzz) which is implementing the same algorithm but is based on a version of fuzzywuzzy that was MIT Licensed and is therefor MIT Licensed aswell, so it can be used in here without forcing a License change. As a nice bonus it is fully implemented in C++ and comes with a few Algorithmic improvements making it between 5 and 100 times faster than FuzzyWuzzy. (actually even faster since python-Levenshtein was not used)

### DIFF
--- a/anime_downloader/sites/twistmoe.py
+++ b/anime_downloader/sites/twistmoe.py
@@ -15,7 +15,7 @@ logger = logging.getLogger(__name__)
 # Don't warn if not using fuzzywuzzy[speedup]
 with warnings.catch_warnings():
     warnings.simplefilter('ignore')
-    from fuzzywuzzy import process
+    from rapidfuzz import process
 
 BLOCK_SIZE = 16
 KEY = b"LXgIVP&PorO68Rq7dTx8N^lP!Fa5sGJ^*XK"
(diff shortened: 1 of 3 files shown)


vn-ki:
I do not want to impose a C++ build system on my users. Especially since installing C/C++ tool chain on windows is a huge task.

maxbachmann:
@vn-ki there are prebuild wheels for Windows https://pypi.org/project/rapidfuzz/0.2.1/#files

vn-ki:
One of the reason I do not want to swap out a dependency because I do not maintain the brew formula and it would cause more headaches than the fear of someone trying to sue me for misusing the GPL license.

As for the breach of license, this is a tool which downloads anime illegally from websites. I think a breach of license is the least illegal thing we're doing. (I'm not proud of either btw)

I will discuss with the one who maintains the formula and see what can be done.

PR #307, opened 2020-03-23
fixed mp4upload & added options to animesimple provider

Blatzar:
The mp4upload extractor was broken (which is used by animesimple sometimes) so I fixed that using deobfuscate_packed_js


Blatzar:
made a new one to make mergin easier

PR #315, opened 2020-03-28
fixed Kickassanime 

Blatzar:
Kickassanime only worked very rarely before because of an oversight, I fixed that and it should work on 95%+ of all shows.
(It doesn't if the only sources are PINK-BIRD and SAPPHIRE-DUCK, which is only on very new episodes)


anime_downloader/sites/kickass.py
```diff
@@ -47,25 +41,52 @@ def _scrape_metadata(self):
             info = json.loads(re.search(regex,str(soup)).group()+']}')
             self.title = info['name']
             
-class KickAssEpisode(AnimeEpisode, sitename='kickassanime'):
+class KickAssEpisode(AnimeEpisode, sitename='kickass'):
         def _get_sources(self):
```
vn-ki: Can you add a TODO here stating that this should be handled at higher level?

anime_downloader/sites/kickass.py
```diff
@@ -5,9 +5,11 @@
 from anime_downloader.sites.anime import Anime, AnimeEpisode, SearchResult
 from anime_downloader.sites import helpers
 
-class KickAss(Anime, sitename='kickassanime'):
-        sitename = 'kickassanime'
-        url = f'https://{sitename}.rs/search'
+logger = logging.getLogger(__name__)
+
+class KickAss(Anime, sitename='kickass'):
```
vn-ki: Document the config here. Look at other sites(gogo, animeflv?) for example.

Issue #300, opened 2020-03-20
Animepahe doesn't seem to be working


sreekaran16:
**Describe the bug**

```
anime dl throws an error "ERROR: 'anime_title'" when trying to download from animepahe. Other sites seem to be working fine. Error existed when running v4.0.1; Updated to v4.1.0 and the error is still there.
```

**To reproduce**

```
anime dl 'dororo' --provider animepahe -e 2: -xd "{aria2}"
```

<details>
<summary>LOG</summary>

```
anime -ll DEBUG dl 'dororo' --provider animepahe -q 720p -e 2: -xd "{aria2}"
2020-03-21 00:25:26 enryuu anime_downloader.util[18212] INFO anime-downloader 4.1.0
2020-03-21 00:25:26 enryuu anime_downloader.util[18212] DEBUG Platform: Windows-10-10.0.18362-SP0
2020-03-21 00:25:26 enryuu anime_downloader.util[18212] DEBUG Python 3.8.2
2020-03-21 00:25:26 enryuu anime_downloader.sites.helpers.request[18212] DEBUG HTML file temp_dir: C:\Users\SKARAN~1\AppData\Local\Temp\animedlox2qc2jp
2020-03-21 00:25:26 enryuu anime_downloader.sites.helpers.request[18212] DEBUG -----
2020-03-21 00:25:26 enryuu anime_downloader.sites.helpers.request[18212] DEBUG GET https://animepahe.com/api
2020-03-21 00:25:26 enryuu anime_downloader.sites.helpers.request[18212] DEBUG {'params': {'l': 8, 'm': 'search', 'q': "'dororo'"}}
2020-03-21 00:25:26 enryuu anime_downloader.sites.helpers.request[18212] DEBUG {'user-agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_7_3) AppleWebKit/535.22 (KHTML, like Gecko) Chrome/19.0.1047.0 Safari/535.22'}
2020-03-21 00:25:26 enryuu anime_downloader.sites.helpers.request[18212] DEBUG -----
[...]

PR #290, opened 2020-03-10
Fixed animepahe

Blatzar:
Fixed animepahe with the help of jsbeautifier, it's not necessary as you can try to deobfuscate the javascript with less LOC or just brute force all possible tokens (8-10), but this was the easiest and cleanest solution.

PR #316, opened 2020-03-28
Fixes AnimePahe and Kwik

ngomile:
animepahe updated the way its API is used. I've updated animepahe to account for this and also made some minor fixes and code clean up here and there. This should #300 but now the remaining thing is kwik which hopefully #290 handles.


vn-ki:
I heard that kwik added captcha(?) This means animepahe is dead right? Pinging @Blatzar

Blatzar:
Yes, there's a captcha on the download page right now. The only way to extract links now is to crack the embedded video player, since there's no captchas there (it seems), not something I can do.

ngomile:
Not quite, the default behavior of the kwik extractor replaces the e in the url with f. So https://kwik.cx/e/n9SsxBkUPN95 becomes https://kwik.cx/f/n9SsxBkUPN95 but as you are about to see in the screenshot if we keep the e it responds correctly.

Capture

This minor tweak should rectify the problem, the captcha doesn't seem to be there anymore either. You can see if the html from the request matches with the kwik page in your browser and from what I saw it does.

Blatzar:
Woah, guess I'll look into it once I solved all my pull requests

Blatzar:
using kwik/e/ I can get an m3u file which mpv fails to play (probably due to referrers)
my kwik.py:
```
import logging
import re
from anime_downloader.extractors.base_extractor import BaseExtractor
import anime_downloader.util as util
logger = logging.getLogger(__name__)

class Kwik(BaseExtractor):
    def _get_data(self):
        print(self.url)
        site = (helpers.get(self.url, referer = self.url, cf = True))
        regex = r"(eval[\W\w]*?)<\/script>"
        javascript = re.search(regex,site.text).group(1)
        deobfuscated = util.deobfuscate_packed_js(javascript)
        regex = r"const source='([^']*)"
        stream = re.search(regex,deobfuscated).group(1)
        return {
            'stream_url': stream,
            'referer': self.url
        }
```

ngomile:
But that confirms it is still working yes?

Blatzar:
Working (ish), What's left to figure out is how to get the m3u file to a playable mp4. (If the m3u file is even correct, I haven't done much testing on it yet)

ngomile:
I don't know much about m3u files either but from doing some Google searches I arrived at [this](https://en.wikipedia.org/wiki/M3U). So from what I have read, perhaps opening it and looking at the contents will lead to the actual url though I am not sure.

ngomile:
Animepahe is likely a lost cause, closing pull request.

vn-ki:
@ngomile look like cloudscraper (https://pypi.org/project/cloudscraper/) works. Can you modify anime-dl to use that and try animepahe again?

ngomile:
Are you using the latest version of Python with it? I'm kind of outdated so requesting https://kwik.cx/e/4tcRdUEckfJx with referer being the same url is still leading to a captcha error. So I'm thinking of updating my Python to see if it makes a difference.

vn-ki:
I didn't specifically tried kwik. Can you try cloudscraper outside of anime dl and see if that works?

ngomile:
Capture
I tried it out and this is what happened, if anyone has a newer installation of Python they should test it and see what happens. I'm a little reluctant to upgrade at the moment.

uali6981:
> Capture
> I tried it out and this is what happened, if anyone has a newer installation of Python they should test it and see what happens. I'm a little reluctant to upgrade at the moment.

Last time when I check this out I also got this error. I didn't check now because didn't have laptop at this time. On that time I thought some dependencies error becuse on `cfscrape` when you check version the `reqiured by` field fill by `anime-downloader` but when I replace this by `cloudscraper` and check `cloudscraper` version I got that field empty.

Blatzar:
I'm having the same issue on Python 3.8.2, however cloudscraper can successfully scrape other sites like kissanime.ru without problems.

ngomile:
I didn't try using the recaptcha option though, seems some need to be paid for and it would probably just add bloat. So kwik is back to square one I guess.

ngomile:
@vn-ki I've fixed kwik and pretty much this is a done deal now.

PR #329, opened 2020-04-08
Fixes to Dreamanime and Ryuanime

AbdullahM0hamed:
Made ryuanime and dreamanime use different server if config server isn't found.

Added xstreamcdn extractor - some of the newer videos only seem to have this server. 

Removed eval from dreamanime


Blatzar:
Please add either fallback or a graceful exit when no episodes in the select language is found. For example when downloading https://www4.ryuanime.com/anime/2700-naruto it throws an error when downloading subbed.

AbdullahM0hamed:
@Blatzar done.

PR #349, opened 2020-04-19
Write Stream Urls to File

ngomile:
This lets users write stream urls to a file, continuously adding on other urls to said file when more calls are made. So if you want Boruto and a bunch of other anime stream urls in a single file this will allow that.


Blatzar:
The anime.txt doesn't respect download directory, it always downloads in current directory.
Something like:
```
        if write:
            util.write_episode(episode,os.path.abspath(download_dir))
```
and
```
def write_episode(episode,path=''):
    # Writes episode stream url to file, further calls
    # append stream urls to file
    make_dir(path)
    with open(f'{path}/anime.txt', 'a+') as f:
        f.write(f'{episode.source().stream_url}\n')
```
Would probably work

ngomile:
All yeah I get you, but when the path is an empty string like that and you try to write to '/anime.txt' it gives an error. So when it's an empty string I'll just make it so that it's set to the current working directory.

ngomile:
I've made changes to prioritize the users specified download directory, links will be written there.

PR #350, opened 2020-04-26
Scrape AnimeFillerList and skip filler episodes

Chr1st-oo:
Ability to be able to skip downloading of filler episodes given an [AnimeFillerList](https://www.animefillerlist.com/) url. 

Example for One Piece, ```https://www.animefillerlist.com/shows/one-piece```

Afaik, AFL has no API so you need to manually enter the url.


Blatzar:
Great PR, I'll add a search function.

Blatzar:
man it took a while but here it is
https://github.com/Chr1st-oo/anime-downloader/pull/1
also fixed a bug that your code didn't skip some episodes e.g 100-150 would only skip 100 and 150

PR #352, opened 2020-04-30
nyaa.si site added

Blatzar:
huge thanks to https://github.com/MichaelDavidJohnson and https://github.com/arjixgamer for all help with porting nyaascrape from https://github.com/MichaelDavidJohnson/NyaaScrape

by default torrents are downloaded using the default program instead of aria, can be changed in config
quality/episode selction doesn't change the search parameters

### DIFF
--- a/anime_downloader/util.py
+++ b/anime_downloader/util.py
@@ -16,7 +16,6 @@ from tabulate import tabulate
 from anime_downloader import session
 from anime_downloader.sites import get_anime_class, helpers
 from anime_downloader.const import desktop_headers
-from anime_downloader.config import Config
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


Blatzar:
BUG: makes config file generation break because of a circular import

anime_downloader/util.py
```diff
@@ -194,6 +194,10 @@ def format_filename(filename, episode):
 
 
 def format_command(cmd, episode, file_format, path):
+    from anime_downloader.config import Config
+    if not Config._CONFIG['dl']['aria2c_for_torrents'] and episode.url.startswith('magnet:?xt=urn:btih:'):
```
vn-ki, line 198: i don't like this at all, but i can't think of an easy workaround for this right now either

PR #360, opened 2020-05-04
added animedaisuki

Blatzar:
openload and streamango are dead it seems so I've 'disabled' them.

### DIFF
--- a/anime_downloader/sites/animedaisuki.py
+++ b/anime_downloader/sites/animedaisuki.py
@@ -1,0 +1,62 @@
+import logging
+import re
+
+from anime_downloader.sites.anime import Anime, AnimeEpisode, SearchResult
+from anime_downloader.sites import helpers
+
+logger = logging.getLogger(__name__)
+
+class Animedaisuki(Anime, sitename='animedaisuki'):
+        sitename = 'animedaisuki'
+        url = f'https://{sitename}.moe/browse'
+        @classmethod
+        def search(cls, query):
+            search_results = helpers.soupify(helpers.get(cls.url, params={'q': query})).select('article > a')
+            search_results = [
+                SearchResult(
+                    title=a.select('h3')[0].text,
+                    url='https://animedaisuki.moe' + a.get('href'))
+                for a in search_results
+            ]
+            return(search_results)
+
+        def _scrape_episodes(self):
+            soup = helpers.soupify(helpers.get(self.url))
+            elements = soup.select('li.fa-play-circle > a')[::-1]
+            print(['https://animedaisuki.moe' + a.get('href') for a in elements if a.get('href').startswith('/watch/')])
+            return ['https://animedaisuki.moe' + a.get('href') for a in elements if a.get('href').startswith('/watch/')]
+
+        def _scrape_metadata(self):
+            soup = helpers.soupify(helpers.get(self.url))
+            self.title = soup.select('h2.Title')[0].text
+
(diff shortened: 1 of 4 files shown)


anime_downloader/sites/animedaisuki.py
```diff
@@ -0,0 +1,62 @@
...
+
+            for a in soup: #removes capcha from link
+                if re.search(captcha_regex,a.get('href')):
+                    links.append(re.search(captcha_regex,a.get('href')).group(1))
+                else:
+                    links.append(a.get('href'))
+            
+            logger.debug(links)
+
+            for a in servers:
```
vn-ki, line 55: Can you actually add this logic to the base class in a later PR? All your providers seem to use something like this.

anime_downloader/sites/animedaisuki.py
```diff
@@ -0,0 +1,62 @@
...
+                    title=a.select('h3')[0].text,
+                    url='https://animedaisuki.moe' + a.get('href'))
+                for a in search_results
+            ]
+            return(search_results)
+
+        def _scrape_episodes(self):
+            soup = helpers.soupify(helpers.get(self.url))
+            elements = soup.select('li.fa-play-circle > a')[::-1]
+            print(['https://animedaisuki.moe' + a.get('href') for a in elements if a.get('href').startswith('/watch/')])
```
vn-ki, line 26: don't print

Blatzar: oopsie, accidentally forgot to remove

anime_downloader/sites/animedaisuki.py
Blatzar: Good idea! Sure will do

Blatzar:
I'm leaving the merge to you as I clearly misunderstood your opinion the last time regarding nyaa

Blatzar:
Regarding the base class for getting the correct server based on config I was thinking something like this:
(reworking anime.py get_data)
```
    def get_data(self):
        self._sources = self._get_sources()

        if type(self._sources[0][0]) != str:

            version = self.config.get('version','subbed') #TODO add a flag for this
            servers = self.config.get('servers',[''])
            data = self._sources[0]
            sources = ''

            logger.debug(data)
            

            for a in servers: #trying all supported servers in order using the correct language
                for b in data:
                    if b['version'] == version:
                        if b['config'] == a:
                            if get_extractor(b['extractor']) == None:
                                continue
                            sources = [(b['extractor'], b['url'],)]
            
            logger.debug('No servers found in selected language. Trying all supported servers')
            
            if not sources:
                for a in servers: #trying all supported servers in order
                    for b in data:
                        if b['config'] == a:
                            if get_extractor(b['extractor']) == None:
                                continue
                            sources = [(b['extractor'], b['url'],)]            
            self._sources = sources

        logger.debug('Sources : {}'.format(self._sources))
```
This way it's still backward compatible (which we later can deprecate) and supports both language and priority of servers through config. For reference, data variable would look something like this:
```
[{'extractor': 'no_extractor', 'url': 'https://animedaisuki.moe/f/Yakusoku no Neverland - 01 [720p]-808d589c.mp4', 'config': 'official', 'version': 'subbed'}, {'extractor': 'no_extractor', 'url': 'https://animedaisuki.moe/f/Yakusoku no Neverland - 01 [720p]-808d589c.mp4', 'config': 'official', 'version': 'subbed'}]
```

PR #362, opened 2020-05-13
Added 2 extractors and reworked vidstream extractor

Blatzar:
Was unsure where to put the extractor config. Maybe adding "extractorconfig" instead?

### DIFF
--- a/anime_downloader/config.py
+++ b/anime_downloader/config.py
@@ -82,6 +82,9 @@ DEFAULT_CONFIG = {
         },
         'animedaisuki': {
             "servers": ["official"]
+        },
+        'vidstream': {
+            "servers": ["vidstream","gcloud","mp4upload","cloud9","hydrax"]
         }
     }
 }
--- a/anime_downloader/extractors/vidstream.py
+++ b/anime_downloader/extractors/vidstream.py
@@ -21,9 +21,9 @@ class VidStream(BaseExtractor):
 
         url = self.url.replace('https:////','https://')
         url = url.replace('https://vidstreaming.io/download','https://vidstreaming.io/server.php')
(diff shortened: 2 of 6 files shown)


anime_downloader/config.py
```diff
@@ -82,6 +82,9 @@
         },
         'animedaisuki': {
             "servers": ["official"]
+        },
+        'vidstream': {
+            "servers": ["vidstream","gcloud","mp4upload","cloud9","hydrax"]
```
vn-ki, line 87: I don't like. Why are these needed?

anime_downloader/extractors/vidstream.py
```diff
@@ -1,42 +1,70 @@
...
 
 class VidStream(BaseExtractor):
     def _get_data(self):
-        QUALITIES = {
-        "360":[],
-        "480":[],
-        "720":[],
-        "1080":[],
+
+        links = {
```
vn-ki, line 15: Are all these supported by the same logic? You could just subclass and add different classes, if thats the case

anime_downloader/config.py
Blatzar: So you could choose which vidstream server you prefer, they're different in terms of download speeds (and probably sizes + quality)

anime_downloader/extractors/vidstream.py
Blatzar: It just links urls to extractors, I don't see any need for classes.

anime_downloader/config.py
vn-ki: You should add this in the docs somewhere and it needs doc comments in the vidstream class.

PR #373, opened 2020-05-25
Added Yts.mx (Yify) as a provider.

ArjixGamer:
This provider is for torrents. For now the quality is only 720p, i plan to change that ASAP.
For some reason the provider has errors, cause it is a magnet link instead of a normal one. When that is fixed the provider will work fine.

### DIFF
--- a/anime_downloader/sites/yify.py
+++ b/anime_downloader/sites/yify.py
@@ -4,7 +4,7 @@ from anime_downloader.sites import helpers
 from bs4 import BeautifulSoup
 from io import StringIO
 
-class Yify(Anime, sitename = 'yify'):
+class Yify(Anime, sitename = 'ytx'):
     sitename = 'yify'
     url = f'https://{sitename}.mx'
     @classmethod
--- a/anime_downloader/sites/yify.py
+++ b/anime_downloader/sites/yify.py
@@ -5,7 +5,7 @@ from bs4 import BeautifulSoup
 from io import StringIO
 
 class Yify(Anime, sitename = 'ytx'):
-    sitename = 'yify'
+    sitename = 'ytx'
     url = f'https://{sitename}.mx'
     @classmethod
     def search(cls, query):
--- a/anime_downloader/sites/yify.py
+++ b/anime_downloader/sites/yify.py
@@ -4,7 +4,7 @@ from anime_downloader.sites import helpers
 from bs4 import BeautifulSoup
 from io import StringIO
 
-class Yify(Anime, sitename = 'ytx'):
+class Yify(Anime, sitename = 'yify'):
     sitename = 'ytx'
     url = f'https://{sitename}.mx'
     @classmethod
--- a/anime_downloader/sites/yify.py
+++ b/anime_downloader/sites/yify.py
@@ -1,0 +1,45 @@
+import requests, json, os
+from anime_downloader.sites.anime import Anime, AnimeEpisode, SearchResult
+from anime_downloader.sites import helpers
+from bs4 import BeautifulSoup
+from io import StringIO
+
+class Yify(Anime, sitename = 'yify'):
+    sitename = 'yify'
+    url = f'https://{sitename}.mx'
+    @classmethod
+    def search(cls, query):
+        search_results = requests.get(f'https://yts.mx/ajax/search?query={query}').text
+        io = StringIO(search_results)
+        results = json.load(io)
+        results_list = []
+        for movie in results['data']:
+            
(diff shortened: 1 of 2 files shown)


anime_downloader/sites/yify.py
```diff
@@ -0,0 +1,45 @@
...
+        io = StringIO(search_results)
+        results = json.load(io)
+        results_list = []
+        for movie in results['data']:
+            
+            """
+            TODO: Find a way to make quality selesction possible.
+                  720p is 'modal-quality-720p'
+                  1080p is 'modal-quality-1080p'
+                  For now the default value is 720p.
```
vn-ki, line 22: Why not use `self.quality`?

anime_downloader/sites/yify.py
```diff
@@ -0,0 +1,45 @@
...
+            """
+            html = BeautifulSoup(requests.get(movie['url']).text, 'html.parser')
```
vn-ki, line 24: This is wrong. You should return `SearchResult`s here and scrape the data in `_scrape_episodes`

anime_downloader/sites/yify.py
```diff
@@ -0,0 +1,45 @@
...
+from anime_downloader.sites import helpers
+from bs4 import BeautifulSoup
+from io import StringIO
+
+class Yify(Anime, sitename = 'yify'):
+    sitename = 'yify'
+    url = f'https://{sitename}.mx'
+    @classmethod
+    def search(cls, query):
+        search_results = requests.get(f'https://yts.mx/ajax/search?query={query}').text
```
vn-ki, line 12: Use `helpers.get` and `helpers.soupify` instead of using `requests` and bs4 directly

anime_downloader/sites/yify.py
```diff
@@ -0,0 +1,45 @@
+import requests, json, os
+from anime_downloader.sites.anime import Anime, AnimeEpisode, SearchResult
+from anime_downloader.sites import helpers
+from bs4 import BeautifulSoup
+from io import StringIO
+
+class Yify(Anime, sitename = 'yify'):
+    sitename = 'yify'
+    url = f'https://{sitename}.mx'
```
AbdullahM0hamed, line 9: yts.mx and yify.mx, although seemingly affiliated are two different sites - could you change the title of your PR to reflect that?

anime_downloader/sites/yify.py
```diff
@@ -0,0 +1,45 @@
...
+            SearchResult(
+                title = results_list[i][0],
+                url = results_list[i][1],
+                meta= results_list[i][2])
+            for i in range(len(results_list))
+            ]
+        return search_results
+        
+    def _scrape_episodes(self):
+        return [self.url] #it is a magnet link, there is nothing else to change
```
AbdullahM0hamed, line 41: Also, yify.mx has TV shows that have episodes (such as this: https://yify.mx/watch/party-of-five-season-6-11HOo/watching.html, which was a random example I found) so you'll want to handle that.

Blatzar:
2 things you NEED to fix to make the provider not give any errors (which it does currently).
1. Move data scraping to _scrape_episodes, currently gives error because site can not be found from the magnet itself (miracle it works on nyaa since there's nyaa in the magnet links)
2. Rename sitename and such, if sitename = yify and url is yts.mx it will give errors

ArjixGamer:
> 2 things you NEED to fix to make the provider not give any errors (which it does currently).
> 
> 1. Move data scraping to _scrape_episodes, currently gives error because site can not be found from the magnet itself (miracle it works on nyaa since there's nyaa in the magnet links)
> 2. Rename sitename and such, if sitename = yify and url is yts.mx it will give errors

thx for noting it, i will fix it as soon as posible. Any ideas on the quality select?

anime_downloader/sites/yify.py
ArjixGamer: dunno why,but when i did use it i got that self is not defined.

Blatzar:
> thx for noting it, i will fix it as soon as posible. Any ideas on the quality select?

Use `self.quality` :|

anime_downloader/sites/yify.py
AbdullahM0hamed: It's because you should be using it in the `_scrape_episodes` method where `self` is the first parameter.

nate-moo:
RIP in peace this provider, should prob close this btw

MichaelDavidJohnson:
Yes. Please close this.

PR #392, opened 2020-06-01
Configurable naming scheme for downloaded files

AbdullahM0hamed:
Because_Not_All_Of_Us_Like_This_Naming_Scheme.

### DIFF
--- a/anime_downloader/util.py
+++ b/anime_downloader/util.py
@@ -16,7 +16,6 @@ from tabulate import tabulate
 from anime_downloader import session
 from anime_downloader.sites import get_anime_class, helpers
 from anime_downloader.const import desktop_headers
-from anime_downloader.config import Config
 
 logger = logging.getLogger(__name__)
 
@@ -176,7 +175,8 @@ def get_json(url, params=None):
     return data
 
 
-def slugify(file_name):
+def slugify(file_name): 
+    from anime_downloader.config import Config
     seperator = Config._CONFIG["dl"]["title_seperator"]
     file_name = str(file_name).strip().replace(' ', seperator)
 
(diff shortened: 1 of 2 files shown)


anime_downloader/util.py
```diff
@@ -16,6 +16,7 @@
 from anime_downloader import session
 from anime_downloader.sites import get_anime_class, helpers
 from anime_downloader.const import desktop_headers
+from anime_downloader.config import Config
```
Blatzar, line 19: Importing config in utils causes circular import on generation of config, moving this to slugify works

ArjixGamer:
dead pr

MichaelDavidJohnson:
Close please.

ArjixGamer:
no, dont close it, dont listen to iggy, he dumb

anime_downloader/util.py
```diff
@@ -175,9 +175,12 @@ def get_json(url, params=None):
     return data
 
 
-def slugify(file_name):
-    file_name = str(file_name).strip().replace(' ', '_')
-    return re.sub(r'(?u)[^-\w.]', '', file_name)
+def slugify(file_name): 
+    from anime_downloader.config import Config
+    seperator = Config._CONFIG["dl"]["title_seperator"]
+    file_name = str(file_name).strip().replace(' ', seperator)
```
vn-ki, line 181: Can you add a TODO to investigate a more robust solution?

Blatzar:
Doesn't work on twist as it retuns names separated with '-' (eg boruto-naruto-next-generations) as file_name

MichaelDavidJohnson:
Yeah if we are gonna do something like remove all _ we should be able to user define if it's like a _ or a -, for future proofing. 
You could also argue it's outside the scope of the project and a batch script could be run after the download to bulk rename, but that's just me :P

vn-ki:
I actually liked this pr. :P 

I understand everyone might not like `_`.

PR #393, opened 2020-06-04
Make header position configurable

AbdullahM0hamed:
Show some love to us folk who prefer the headers to be at the top.

### DIFF
--- a/anime_downloader/config.py
+++ b/anime_downloader/config.py
@@ -23,6 +23,7 @@ DEFAULT_CONFIG = {
         'selescrape_browser': None,
         'selescrape_browser_executable_path' : None,
         'selescrape_driver_binary_path' : None,
+        'header_position': 'bottom'
     },
     'watch': {
         'quality': '1080p',
--- a/anime_downloader/util.py
+++ b/anime_downloader/util.py
@@ -16,6 +16,7 @@ from tabulate import tabulate
 from anime_downloader import session
 from anime_downloader.sites import get_anime_class, helpers
 from anime_downloader.const import desktop_headers
+from anime_downloader.config import Config
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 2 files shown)


AbdullahM0hamed:
Perhaps you will find this screenshot useful, the top is when the value of 'header_position' is set to 'top', and the bottom is when it is set to bottom (which is the default).

anime_downloader/config.py
```diff
@@ -21,6 +21,9 @@
         'external_downloader': '',
         'aria2c_for_torrents': False,
     },
+    'search': {
```
vn-ki, line 24: don't need a new section. put it in `dl` maybe?

anime_downloader/util.py
```diff
@@ -68,7 +75,11 @@ def format_search_results(search_results):
     table = [(i+1, v.title, v.pretty_metadata)
              for i, v in enumerate(search_results)]
     table = tabulate(table, headers, tablefmt='psql')
-    table = '\n'.join(table.split('\n')[::-1])
+    try:
+        layout = positions[Config._CONFIG["search"]["header_position"]]
+    except KeyError:
+        raise KeyError(f"The value '{Config._CONFIG['search']['header_position']}' is not a valid position for the headers")
```
vn-ki, line 81: Warn, don't error

AbdullahM0hamed:
Done

anime_downloader/util.py
```diff
@@ -68,7 +75,12 @@ def format_search_results(search_results):
     table = [(i+1, v.title, v.pretty_metadata)
              for i, v in enumerate(search_results)]
     table = tabulate(table, headers, tablefmt='psql')
-    table = '\n'.join(table.split('\n')[::-1])
```
vn-ki, line 71: Needs a default `layout`

AbdullahM0hamed: Gave it a default value of -1 on line 79 now.

vn-ki:
Conflicts

AbdullahM0hamed:
I believe that she be solved now

anime_downloader/config.py
```diff
@@ -23,6 +23,7 @@
         'selescrape_browser': None,
         'selescrape_browser_executable_path' : None,
         'selescrape_driver_binary_path' : None,
+        'header_position': 'bottom'
```
vn-ki, line 26: I guess this should be in a new section `common`, but we can do that later.

vn-ki:
Reverted. You have broken the build.

PR #394, opened 2020-06-04
Let user change from dubbed to subbed if no episodes are found - for ryuanime and dreamanime 

AbdullahM0hamed:
Had this idea a long time ago, and I was told I couldn't take input from providers. But apparently I can, so here I am: bringing 2 month old ideas back to life for the sake of it.

### DIFF
--- a/anime_downloader/sites/dreamanime.py
+++ b/anime_downloader/sites/dreamanime.py
@@ -36,10 +36,8 @@ class DreamAnime(Anime, sitename='dreamanime'):
 
         return search_results
 
-    def _scrape_episodes(self, version = None):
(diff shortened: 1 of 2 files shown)


Blatzar:
no option to change from subbed to dubbed :neutral_face:

AbdullahM0hamed:
Probably a good idea to have it just in case. Didn't consider it as a viable possibility, rip

AbdullahM0hamed:
Done

vn-ki:
To explain, this should be handled in `dl.py`. Click absolutely should not be improted in `sites/*` or `extractors/*`

PR #396, opened 2020-06-06
Solve Kwik captcha

AbdullahM0hamed:
This solves @LordShenron's new issue in #375.

The first episode it ever retrieves will take a while (not too long, though), as it is attempting to pass the hcaptcha. The following instances will all be normal speed because cookies are stored for future use.

I've opted to use requests directly instead of helpers for requests, simply because helpers errors out when there are 403 errors (which is guaranteed to happen the first time) - and I did not want to deal with that.

### DIFF
--- a/anime_downloader/extractors/kwik.py
+++ b/anime_downloader/extractors/kwik.py
@@ -28,6 +28,7 @@ class Kwik(BaseExtractor):
                 ))
             }
     session = requests.session()
+    token = ''
 
 
     #Captcha bypass stuff is mostly thanks to https://github.com/Futei/SineCaptcha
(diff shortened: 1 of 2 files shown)


LordShenron:
I just tried with the new kwik extractor and its working.. It did failed for the first try but on the second it bypassed hcaptcha. 
nn

Its better than anything :) 
Thanks for fixing it.

anime_downloader/extractors/kwik.py
```diff
@@ -22,20 +100,40 @@ def _get_data(self):
 
         #Necessary
         self.url = self.url.replace(".cx/e/", ".cx/f/")
+        self.headers.update({"referer": self.url})
+
+        if not os.path.isfile(APP_DIR + '/kwik'):
```
vn-ki, line 105: Why not use temp dir instead of `ARR_DIR`?

anime_downloader/extractors/kwik.py
```diff
@@ -22,20 +100,40 @@ def _get_data(self):
...
+            logger.info("Bypassing captcha...")
+            self.bypass_captcha()
+
+            resp = helpers.soupify(self.session.get(self.url, headers = self.headers))
+            bypass_url = 'https://kwik.cx' + resp.form.get('action')
+
+            data = {}
+            [data.update({x.get("name"): x.get("value")}) for x in resp.select("form > input")]
```
vn-ki, line 113: replace by 
```
data = dict((x.get("name"): x.get("value")) for x in resp.select("form > input"))
```

AbdullahM0hamed:
Dome

PR #399, opened 2020-06-08
Reduce fake mouse movements being sent

AbdullahM0hamed:
Hcaptcha was sending more mouse movement data than they seem to care about, and I naively believed what I saw.

I'm here to rectify my foolish mistakes.

### DIFF
--- a/anime_downloader/extractors/kwik.py
+++ b/anime_downloader/extractors/kwik.py
@@ -62,10 +62,6 @@ class Kwik(BaseExtractor):
 
             #Mouse movements
             mm = self._generate_mouse_movements(timestamp)
-            ts = self._generate_mouse_movements(timestamp)
-            te = self._generate_mouse_movements(timestamp)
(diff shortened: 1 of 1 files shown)


ngomile:
Have you tried using this solution on vidstream? It might work, could become a useful utility function.

AbdullahM0hamed:
Can't find any vidstream links with hcaptcha at the moment, know of any?

ngomile:
Try this [link to Trinity Blood](https://vidstreaming.io/download?id=MjU1MDY=&typesub=Gogoanime-SUB&title=Trinity+Blood+Episode+1) out. Should immediately show hcaptcha or perhaps its picky.

Blatzar:
CAPTCHAs show on vidstream if you press the download button, not stream. Trying to bypass vidstream CAPTCHAs isn't worth the effort and future problems IMO as it already works.

PR #401, opened 2020-06-08
fixed the logger for selescrape

ArjixGamer:


AbdullahM0hamed:
@ArjixGamer does this supersede #397? If it does, you should go ahead and close that.

PR #402, opened 2020-06-08
Moved hcaptcha and cookie stuff to utils

AbdullahM0hamed:
This adds the hcaptcha bypass to util, along with a `get_hcaptcha_cookies` function. Currently, this is implemented in Kwik only - but works with other sites (tested with vidstreaming link provided by @ngomile, however haven't implemented it due to @Blatzar's comment on #399, which this supersedes).

This is also less error prone as it catches errors that occasionally occur (namely KeyError and TypeError - due to an unexpected or no response from the server on some failures).

### DIFF
--- a/anime_downloader/util.py
+++ b/anime_downloader/util.py
@@ -318,7 +318,7 @@ def bypass_hcaptcha(url):
 def get_hcaptcha_cookies(url):
     """
     :param url: url that you want to use cookies for
-    :return: returns cookies if they were stored, or False, if they weren't
+    :return: returns cookies if they were stored, or nothing, if they weren't
     """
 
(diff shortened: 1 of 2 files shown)


anime_downloader/util.py
```diff
@@ -226,6 +233,99 @@ def format_command(cmd, episode, file_format, path):
     return cmd
 
 
+#Credits to: https://github.com/Futei/SineCaptcha - you should check that out by the way, it's hilarious
```
vn-ki, line 236: remove this
```
you should check that out by the way, it's hilarious
```

AbdullahM0hamed: Done

PR #404, opened 2020-06-09
fixed Kissanime

Blatzar:
Made changes to selescrape and not merging instantly as I'm not very experienced when it comes to classes and the proper way to use them.

Huge thanks to @AbdullahM0hamed for help with both hydrax and kissanime search!

### DIFF
--- a/anime_downloader/sites/helpers/selescrape.py
+++ b/anime_downloader/sites/helpers/selescrape.py
@@ -173,23 +173,54 @@ def cloudflare_wait(driver):
             break
     time.sleep(1) # This is necessary to make sure everything has loaded fine.
 
-class get_session:
-    def request(self, request_type, url, **kwargs): #Headers not yet supported , headers={}
+
+def request(request_type, url, **kwargs): #Headers not yet supported , headers={}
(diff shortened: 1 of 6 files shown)


AbdullahM0hamed:
404: PR not found.

anime_downloader/sites/helpers/selescrape.py
```diff
@@ -170,19 +173,23 @@ def cloudflare_wait(driver):
...
-        cloudflare_wait(driver)
-        html = driver.page_source
-        driver.close()
-        return html
-    except:
-        driver.save_screenshot(f"{get_data_dir()}/screenshot.png");
-        driver.close()
-        logger.error(f'There was a problem getting the page: {new_url}. \
-        See the screenshot for more info:\n{get_data_dir()}/screenshot.png')
+class get_session:
```
vn-ki, line 176: Should be returning a `SeleResponse` class which has the required fields.

Blatzar:
Pretty sure it fails because one (or more) of the following reasons:
Selenium not installed and/or configured by default 
Kissanime tests are outdated (looks like they're using cfscrape)

PR #409, opened 2020-06-14
Make kwik less error prone

AbdullahM0hamed:
People have been experiencing errors with deobfuscating the js returned by Kwik every once in a while but would always work after a retry. This attempts to fix that.


Blatzar:
I'd like to see a max retry limit before I feel comfortable merging this

Issue #418, opened 2020-06-25
List of anime , auto shutdown


shock-wave007:
**Describe the feature here**
if i can rate this library i will give it 10/5 :P
amazing work thanks for this library and keep doing good work and hope ur well in this time of pandemic

1. i can queue multiple anime at the beginning. like find and add anime to download queue.
2. auto shutdown when everything is complete

3. i think iam thinking too far but if possible option to pause

anyway best repo i have ever come across on github

Blatzar:
Number 1 can definitely get added, but I think it requires further rewrite of the core. I think using anime-downloader as a library is the best option for this one. 
Number 2 is very easy to add, can probably make a PR like today or tomorrow. 
Number 3 is already kinda added as aria2 can resume canceled downloads.

PR #419, opened 2020-06-26
added shutdown flag

Blatzar:
Just added a shutdown flag in response to https://github.com/vn-ki/anime-downloader/issues/418. Not sure if redundant as you can always do `&& shutdown`, hence I'm leaving the merging to you @vn-ki

### DIFF
--- a/anime_downloader/commands/dl.py
+++ b/anime_downloader/commands/dl.py
@@ -1,5 +1,6 @@
 import logging
 import os
+import sys
 
 import click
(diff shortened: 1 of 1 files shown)


vn-ki:
We shouldn't be adding this flag tbh. It should be done by an external script. (bash or bat)

ArjixGamer:
how about adding a flag that executes any external script/program upon completion?
that way i can make it hibernate instead

PR #426, opened 2020-07-05
Add dbanimes

AbdullahM0hamed:
Tested, but not extensively. Some episodes only use unsupported servers, which I'll maybe try to support later (like clipwatching - of which I could find no working link, or jetload - which uses invisible Google recaptcha to generate a token that it then posts along with the video id to jetload.net/jet_secure, or vidfast - which simply wouldn't play for me, rip)


anime_downloader/config.py
```diff
@@ -128,6 +128,10 @@
                 "gounlimited"
             ]
         },
+        'dbanimes': {
+            'server': 'fembed',
+            'fallback_servers': ['mixdrop', 'sendvid']
```
Blatzar, line 133: I suggest using sort_sources instead, or at least in the future for less headaches with selected servers and all that.

nate-moo:
`426: Upgrade Required`

Blatzar:
Going to fix the rest

Blatzar:
TODO: Sources found on https://dbanimes.com/your-name-vostfr/ can probably get added

PR #427, opened 2020-07-06
added vidstream as a proper provider

Blatzar:
Better to just get from the source


AbdullahM0hamed:
Bruh, I'm quite baffled that vidstream is an actual anime site😭😭😭😭. I thought it was a hosting site or something, how naive I was.

PR #430, opened 2020-07-09
Added provider animeonline360

ArjixGamer:


ArjixGamer:
plz merge

anime_downloader/extractors/animeonline360.py
```diff
@@ -0,0 +1,18 @@
...
+
+class AnimeOnline360(BaseExtractor):
+    def _get_data(self):
+        try:
+            html = helpers.soupify(helpers.get(self.url))
+            soup = html.find('li', class_='dooplay_player_option')
+            anime_type = soup['data-type']
+            anime_id = soup['data-post']
+            referrer = helpers.get(f"https://animeonline360.me/wp-json/dooplayer/v1/post/{anime_id}?type={anime_type}&source=1").json()['embed_url']
+            stream_url = referrer.split('?source=')[1].split('&')[0].replace('%2F', '/').replace('%3A', ':')
```
Blatzar, line 12: Would personally use regex here, also the way you do .replace() can lead to future issues. Are they necessary? can you use something like urllib

anime_downloader/sites/animeonline360.py
```diff
@@ -0,0 +1,55 @@
+import json, requests
+
+from anime_downloader.sites.anime import Anime, AnimeEpisode, SearchResult
+from anime_downloader.sites import helpers
+
+class AnimeOnline(Anime, sitename = 'animeonline360'):
+
+    sitename = 'animeonline360'
+    url = f'https://{sitename}.me'
```
Blatzar, line 9: probably not needed if you're not using it

ArjixGamer: ok

anime_downloader/extractors/animeonline360.py
ArjixGamer: i can do this instead: https://stackoverflow.com/questions/16566069/url-decode-utf-8-in-python

ArjixGamer: it was just a lazy and fast solution i had, i will use the url decode

ArjixGamer:
wryyyyy

anime_downloader/extractors/animeonline360.py
```diff
@@ -0,0 +1,18 @@
...
+            anime_type = soup['data-type']
+            anime_id = soup['data-post']
+            referrer = helpers.get(f"https://animeonline360.me/wp-json/dooplayer/v1/post/{anime_id}?type={anime_type}&source=1").json()['embed_url']
+            stream_url = urllib.parse.unquote(referrer.split('?source=')[1].split('&')[0])
+            return {
+                'stream_url': stream_url,
+                'referer': referrer
+            }
+        except:
+        	return {"stream_url": ''}
```
AbdullahM0hamed, line 18: This seems to be more indented than it should be.

anime_downloader/sites/animeonline360.py
```diff
@@ -0,0 +1,54 @@
...
+class AnimeOnline(Anime, sitename = 'animeonline360'):
+
+    sitename = 'animeonline360'
+
+    @classmethod
+    def search(cls, query):
+        try:
+            # r = helpers.get('https://animeonline360.me/wp-json/dooplay/search/?nonce=12d75e884b', params = {'keyword': query}).json()
+            r = helpers.soupify(helpers.get('https://animeonline360.me/', params = {'s': query})).select('div.title')
+            results = [{"title": x.text, "url": x.a['href']} for x in r]
```
Blatzar, line 15: This seems redundant, move down one step to SearchResult

anime_downloader/sites/animeonline360.py
```diff
@@ -0,0 +1,54 @@
...
+                    meta= {})
+
+                for i in results
+                ]
+
+            return search_results
+        except:
+            return ""
+
+    def _scrape_episodes(self):
```
Blatzar, line 29: You can replace this entire function with
```
    def _scrape_episodes(self):
        data = helpers.soupify(helpers.get(self.url)).select('div.episodiotitle > a')
        return [i.get('href') for i in data[::-1]]
```
Should work just as fine, no need to make things too complicated

anime_downloader/sites/init.py
```diff
@@ -2,6 +2,7 @@
 
 ALL_ANIME_SITES = [
     # ('filename', 'sitename', 'classname')
+    ('animeonline360','animeonline360','AnimeOnline'),
```
Blatzar, line 5: I'd personally prefer if you removed the 360, it's too long

anime_downloader/extractors/animeonline360.py
ArjixGamer: oh snap

anime_downloader/sites/animeonline360.py
ArjixGamer: can you be more specific? do you mean the episode titles or the commented line of code? if its the latter then i will remove it.

anime_downloader/sites/animeonline360.py
ArjixGamer: ok

anime_downloader/sites/init.py
ArjixGamer: from the filename? ok

anime_downloader/sites/animeonline360.py
Blatzar: Instead of saving it to a variable and then using it you can use it directly in searchreuslts like this. 
```
            search_results = [
                SearchResult(
                    title = i.text,
                    url = i.a['href'],
                    meta= {})
                for i in r
            ]
```

anime_downloader/sites/init.py
Blatzar: sitename, to make `--provider` shorter

anime_downloader/sites/animeonline360.py
ArjixGamer: I see no problem with the way i have written it. So i wont change it.

ArjixGamer: So that i can resolve this comment i will say this:
The code works as intended, it is not bad and it is 100% readable. There is no reason to change it just because you prefer another way to achieve the same exact result. There is no profit either by changing the code.

PR #431, opened 2020-07-10
Adding Anime Frenzy

nate-moo:
is my git history broken? if so im sorry and wil try to fix it. if not then great!

### DIFF
--- a/anime_downloader/config.py
+++ b/anime_downloader/config.py
@@ -31,6 +31,9 @@ DEFAULT_CONFIG = {
         'provider': 'twist.moe',
     },
     "siteconfig": {
+        'animixplay': {
+            "server": "vidstream",
+        },
         'nineanime': {
             "server": "mp4upload",
         },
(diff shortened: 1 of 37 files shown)


nate-moo:
`431: Request Header Fields Too Large`

anime_downloader/config.py
```diff
@@ -31,6 +31,10 @@
         'provider': 'twist.moe',
     },
     "siteconfig": {
+        'animefrenzy': {
+            "lang": "subbed"
+            #Lang Key can either be "subbed" or "dubbed"
```
Blatzar, line 36: I heavily doubt comments in config works

anime_downloader/config.py
```diff
@@ -31,6 +31,10 @@
         'provider': 'twist.moe',
     },
     "siteconfig": {
+        'animefrenzy': {
+            "lang": "subbed"
```
Blatzar, line 35: 'version' is often used for languages, I'd prefer that for consistency

nate-moo:
@Blatzar I removed the comment from config.py, and changed the key from lang to version

PR #441, opened 2020-07-15
MAL implementation

Blatzar:
Right now rough proof of concept for MAL implementation, building on top of this shouldn't be very hard. 

It basically uses the first result on MAL to choose the most similar search result automatically.

Current issues which needs to get adressed:
Extra flags in title needs to get removed in providers to make it more reliable.
Sub/Dub needs to get added to metadata to make that automatic.

This paves the way for fully automatic seamless provider fallback and provides the metadata needed for good GUI.


anime_downloader/animeinfo.py
```diff
@@ -0,0 +1,83 @@
...
+        for j in search_results:
+            # 0 if there's no japanese name
+            jap_ratio = fuzz.ratio(i.jap_title, j.meta_info['jap_title']) if j.meta_info.get('jap_title') else 0
+            # Outputs the max ratio for japanese or english name (0-100)
+            ratio = max(fuzz.ratio(i.title,j.title), jap_ratio)
+            results.append([ratio,j])
+
+    # Returns the result with highest ratio
+    return max(results, key=lambda item:item[0])[1]
+    # Should probably be made an object or dict to include the MAL info
```
Blatzar, line 83: Something like MatchObject

anime_downloader/sites/anime.py
```diff
@@ -453,13 +453,16 @@ class SearchResult:
         URL for the poster of the anime.
     meta: dict
         Additional metadata regarding the anime.
+    meta_info: dict
+        Metadata regarding the anime. Not shown in the results, used to match with MAL
     """
 
-    def __init__(self, title, url, poster='', meta=''):
+    def __init__(self, title, url, poster='', meta='', meta_info={}):
```
Blatzar, line 460: Needed to provide more metadata without impacting the search results showed to the user.

MichaelDavidJohnson:
Not to be that guy, but consider renaming the "jap_*" variables to avoid being cancelled in the future.

AbdullahM0hamed:
441: PR is at / over data quota

Blatzar:
Updated it to include functions for automatic dub select and a cleaned title.

Dub select is a bit ugly atm, but couldn't find a cleaner solution.

The cleaned title is basically the raw title without (2002) (RAW), useful for some sites like animeout with verbose titles which may make the fuzzy matching worse. 

Example sites I used are vidstream and animeout

ArjixGamer:
make it a draft since it is not completely done yet.

Blatzar:
It already is and has been a draft since the start

ArjixGamer:
oh

Blatzar:
Recent changes should make sub/dub choice more reliable and easier to configure. Prev build had a ~~feature~~ bug which chose dubbed even if sub was selected. ~~Slipped by since I always use the superior language~~

anime_downloader/animeinfo.py
```diff
@@ -0,0 +1,134 @@
...
+            # Allows for returning of cleaned title by the provider using 'title_cleaned' in meta_info.
+            # To make fuzzy matching better.
+            # TODO allow this for japanese titles too
+            title_provider = j.title if not j.meta_info.get('title_cleaned') else j.meta_info.get('title_cleaned')
+            title_info = i.title
+
+            # Essentially adds the chosen key to the query if the version is in use
+            # Dirty solution, but should work pretty well
+
+            config = Config['siteconfig'].get(get_anime_class(j.url).sitename,{})
```
Blatzar, line 118: The cleanest way I could access config from here. Config is needed to automatically choose dub or subbed based on title.

Blatzar:
Recent changes is the beginning of making fallback_providers. For anyone who wants to help, check the TODO:s in dl.py

Blatzar:
Now correctly works on an episode basis. However the current implementation with obliterate any selenium providers as they have no cache.

anime_downloader/config.py
```diff
@@ -16,13 +16,14 @@
       	'chunk_size': '10',
         'fallback_qualities': ['720p', '480p', '360p'],
         'force_download': False,
-        'file_format': '{anime_title}/{anime_title}_{ep_no}',
+        'file_format': '{animeinfo_anime_title}/{animeinfo_anime_title}_{provider}_{ep_no}',
```
Blatzar, line 19: Retarded config to show off the function, make this better before merge

Blatzar:
Latest push made choices persistent and fixed downloaded title

Blatzar:
Fixed config based info provider and and exit when all providers are skipped

Blatzar:
TODO: season flag and season_key specified in provider

ArjixGamer:
That's some good progress i must say.

anime_downloader/commands/dl.py
```diff
@@ -42,7 +44,7 @@
     '--force-download', '-f', is_flag=True,
     help='Force downloads even if file exists')
 @click.option(
-    '--file-format', '-ff', default='{anime_title}/{anime_title}_{ep_no}',
+    '--file-format', '-ff', default='{animeinfo_anime_title}/{animeinfo_anime_title}_{provider}_{ep_no}',
```
vn-ki, line 47: I dont think it's a good idea to change the default file format.

What would happen if the anime returned by MAL is the wrong one?

anime_downloader/animeinfo.py
```diff
@@ -0,0 +1,216 @@
...
+
+    # Code below uses the first result to compare
+    #season_info = [scrape_metadata(search_results[0].url)] 
+    #return season_info
+
+    # Prompts the user for selection
+    return primitive_search(season_info)
+
+
+def search_anilist(query):
```
vn-ki, line 125: What is the use of anilist search if it's doing the exact same thing as MAL?

anime_downloader/commands/dl.py
Blatzar: Merely feature showcase, will fix before merge

anime_downloader/animeinfo.py
Blatzar: Will remove MAL

Blatzar:
Currently very slow compared to dl, investigate

Blatzar:
(very) Slow only on twist.moe (it seems, havent tried all providers). 
image
This is probably due to the fuzzywuzzy.process.extract(query, animes) getting run excessivley which only really concerns twist.

Some speed issues can be found with normal providers, but 2.5s compared to 0.5 on 12 cached episodes (instead of 10s vs 1s on twist). This is caused by helpers.soupify() which is acceptable in my eyes. 

Some optimization can be done by running all episodes at once one provider at a time (like current dl, but with a for loop for providers, which can be threaded) and then comparing results for lacking episodes/errors. This mainly comes at the cost of sphagetti code, but is completely doable.

If you want @vn-ki I'll start working on that.

anime_downloader/commands/ezdl.py
```diff
@@ -0,0 +1,202 @@
...
+    '--choice', '-c',type=int,
+    help='Choice to start downloading given anime number '
+)
+@click.option(
+    '--fallback-providers', '-fp', cls=util.ClickListOption,
+    help='Specifiy the order of fallback providers as a list.')
+@click.option("--skip-fillers", is_flag=True, help="Skip downloading of fillers.")
+
+@click.pass_context
+def command(ctx, anime_url, episode_range, url, player, skip_download, quality,
```
vn-ki, line 82: Some of the options can be removed for `ezdl` to make it easier?

anime_downloader/commands/ezdl.py
```diff
@@ -0,0 +1,202 @@
...
+    # Stores the choices for each provider, to prevent re-prompting search.
+    choice_dict = {}
+
+    for episode_range in range(int(episode_range_split[0]), int(episode_range_split[-1])+1):
+        # Exits if all providers are skipped.
+        if [choice_dict[i] for i in choice_dict] == [0]*len(providers):
+            logger.info('All providers skipped, exiting')
+            exit()
+
+        for provider in providers:
```
vn-ki, line 111: Why check provider for every episode?

anime_downloader/commands/ezdl.py
```diff
@@ -0,0 +1,202 @@
...
+    episode_count = info.episodes - 1
+    # Interprets the episode range for use in a for loop.
+    # 1:3 -> for _episode in range(1, 4):
+    episode_range = util.parse_episode_range(episode_count, episode_range)
+    episode_range_split = episode_range.split(':')
+
+    # Stores the choices for each provider, to prevent re-prompting search.
+    choice_dict = {}
+
+    for episode_range in range(int(episode_range_split[0]), int(episode_range_split[-1])+1):
```
vn-ki, line 105: What exactly is happening here?

anime_downloader/commands/ezdl.py
Blatzar: Let's say twist was missing only ep 4 on a show. To make only that episode fall back you'd either do this or run all the providers and compare results.

anime_downloader/commands/ezdl.py
Blatzar: See line 96-100 for more context. What it does is convert the episode range to a for loop using that range correctly.

anime_downloader/commands/ezdl.py
Blatzar: See line 96-100 for more context. What it does is convert the episode range to a for loop using that range correctly.

Blatzar: See line 96-100 for more context. What it does is convert the episode range to a for loop using that range correctly.

Blatzar:
Bug: search() returning 2 values will mess with watch, I'd need to rebase before merging to prevent errors

Blatzar:
Closing as I rebased

PR #443, opened 2020-07-20
Replace Pycryptodome

nate-moo:
uses pyaes instead of pycryptodome for full compatibility


nate-moo:
Stopped working

nate-moo:
ok, so the encrypted ciphertext length is 32 but it needs to be 16

nate-moo:
ready for testing and 90% sure it works

vn-ki:
This will break things for AUR and brew users.

vn-ki:
Also looks like pyaes package isn't maintained anymore (last update 3 years ago).

MichaelDavidJohnson:
Yeah I think the best way to deal with this is just making an ios fork, the question is should we add the fork as a fork of anime-dl?

PR #447, opened 2020-07-23
Added referer and window title to mpv

ArjixGamer:
as the title says i added a referer and window title to mpv so that people like me that have rich presence for mpv can showoff what they are watching instead of having it show ``playing video.mp4``


ArjixGamer:
plz merge

MichaelDavidJohnson:
As long as it works for all OS there's no reason not to merge this imo

Blatzar:
Looks good, will do some testing later

nate-moo:
> plz merge

*no*

PR #448, opened 2020-07-24
Re:upload Update command

ArjixGamer:
I dont remember how or why i deleted the old branch so the pr close. here is a reupload/rework of it


Blatzar:
Bloat and unnessecary crust which doesn't belong in this project. Users will have to use pip install --upgrade like any other python program.

PR #457, opened 2020-07-28
Added provider 4anime.to

ArjixGamer:


ArjixGamer:
it would be smaller in loc if i didnt add year metadata for it, but i think its worth it

ArjixGamer:
oh yeah the README.md

ArjixGamer:
> Looks good from reading it, will do testing later, add the site to README

i tried following your past advice from previous PRs so that i dont have to change much for it to get merged

anime_downloader/sites/4anime.py
```diff
@@ -0,0 +1,45 @@
...
+        return [x.a['href'] for x in soup]
+
+    def _scrape_metadata(self):
+        soup = helpers.soupify(helpers.get(self.url).text)
+        self.title = soup.title.text
+        for item in soup.find_all('div', class_='detail'):
+            more = item.find_all('a')
+            for i in more:
+                if 'year' in i['href']:
+                    self.year = int(i.text)
```
Blatzar, line 40: honestly no idea if this is a standard or should be done líke this

ArjixGamer: it was the only way i could get the year without using an index number that would be prone to errors

ArjixGamer: it might be an overkill just because of the Lines Of Code it consumes, but it only iterates over 5 or 6 items and all the links in each of item, not much to fear for slowdown

Blatzar: No faults in the code, just unsure if this is how to correctly define metadata in this project

ArjixGamer: by define you meant how i set self.year?

Blatzar: yes, it could be something like self.metadata['year'] = 2020 for all I know.

ArjixGamer: > yes, it could be something like self.metadata['year'] = 2020 for all I know.
if we decide on a standard for the metadata in __scrape_metadata then we can easily change how it gets saved as metadata

anime_downloader/sites/4anime.py
```diff
@@ -0,0 +1,44 @@
...
+    def _scrape_episodes(self):
+        soup = helpers.soupify(helpers.get(self.url).text).find('ul', class_='episodes range active').find_all('li')
+        return [x.a['href'] for x in soup]
+
+    def _scrape_metadata(self):
+        soup = helpers.soupify(helpers.get(self.url).text)
+        self.title = soup.title.text
+        for i in soup.select('.detail > a'):
+            if 'year' in i.get('href',''):
+                self.meta = {}
```
Blatzar, line 38: No need to define self.meta as it's already defined in anime.py

anime_downloader/sites/4anime.py
```diff
@@ -0,0 +1,44 @@
...
+            for i in soup.find_all('div', class_='item')
+            ]
+        return search_results
+
+    def _scrape_episodes(self):
+        soup = helpers.soupify(helpers.get(self.url).text).find('ul', class_='episodes range active').find_all('li')
+        return [x.a['href'] for x in soup]
+
+    def _scrape_metadata(self):
+        soup = helpers.soupify(helpers.get(self.url).text)
```
Blatzar, line 34: .text not necessary

Blatzar: same goes for L30, L18. Soupify solves that

anime_downloader/sites/4anime.py
```diff
@@ -0,0 +1,44 @@
...
+        self.title = soup.title.text
+        for i in soup.select('.detail > a'):
+            if 'year' in i.get('href',''):
+                self.meta = {}
+                self.meta['year'] = int(i.text) if i.text.isnumeric() else None
+
+class Anime4Episode(AnimeEpisode, sitename='4anime'):
+    def _get_sources(self):
+        stream_url = helpers.soupify(helpers.get(self.url).text).find('div', class_='videojs-desktop').find('source')['src']
```
Blatzar, line 43: .select("source") also works

anime_downloader/sites/4anime.py
```diff
@@ -0,0 +1,44 @@
...
+
+    @classmethod
+    def search(cls, query):
+        data = {
+            "action": "ajaxsearchlite_search", 
+            "aslp": query, 
+            "asid": 1, 
+            "options": "qtranslate_lang=0&set_intitle=None&customset%5B%5D=anime"
+            }
+        soup = helpers.soupify(helpers.post("https://4anime.to/wp-admin/admin-ajax.php", data=data).text)
```
Blatzar, line 18: This can be simplified to 
```
        search_results = helpers.soupify(helpers.post("https://4anime.to/wp-admin/admin-ajax.php", data=data)).select('div.info > a')

        search_results = [
            SearchResult(
                title = i.text,
                url = i['href']
                )
            for i in search_results
            ]
        return search_results
```

ArjixGamer: nice

anime_downloader/sites/4anime.py
ArjixGamer: ok, ill keep that in mind, thx

anime_downloader/sites/4anime.py
```diff
@@ -0,0 +1,44 @@
...
+            SearchResult(
+                title = i.find('div', class_='info').a.text,
+                url = i.find('div', class_='info').a['href']
+                )
+            for i in soup.find_all('div', class_='item')
+            ]
+        return search_results
+
+    def _scrape_episodes(self):
+        soup = helpers.soupify(helpers.get(self.url).text).find('ul', class_='episodes range active').find_all('li')
```
Blatzar, line 30: Can get changed to 
```
        soup = helpers.soupify(helpers.get(self.url)).select('ul.episodes.range.active > li > a')
        return [i['href'] for i in soup]
```
Use css selectors for easier and more readable selection

anime_downloader/sites/4anime.py
ArjixGamer: last time i didnt use .text i had a lot of errors

anime_downloader/sites/4anime.py
ArjixGamer: oh ok

PR #458, opened 2020-07-28
fixed banner image

ArjixGamer:
i dont know when/how or why you removed my banner image link that i uploaded to imgur, but here is it. i hope you merge it


MichaelDavidJohnson:
Why do you have to sound so passive aggressive, it could have just been an error

ArjixGamer:
> Why do you have to sound so passive aggressive, it could have just been an error

cause it is annoying, maybe somebody used git console to push all the files and had the old readme.md

PR #459, opened 2020-07-31
Change iOS installation instructions to install from my fork.

nate-moo:


Blatzar:
I'm no fan of users relying on an unofficial potentiality outdated fork. Better trying to integrate it with the current twist imo

AbdullahM0hamed:
@Blatzar, *sighs* - you say that as if he hasn't already done that, and if his implementation wasn't rejected because of how it would affect Mac users. It's either an outdated fork, or they don't use it at all, unfortunately. Unless the instructions are changed to make them git clone the branch, manually edit setup.py, and pip install that.

nate-moo:
@blatzar would you like me to keep the old instructions as well as add the new ones in case whomever trys this (if anyone does) wants the bleeding edge without waiting for me to update my branch? 
Red Guy was referring to this btw: https://github.com/vn-ki/anime-downloader/pull/443#issuecomment-662447368

MichaelDavidJohnson:
I'm not sure on this. I know where you're coming from but I can't see a way to really advertise your fork. It could be outdated, pip upgrade wont work ect... 

Maybe there's a smart way to merge the fork inside a fork in anime-dl that keeps that functionality?

PR #460, opened 2020-08-01
added wget as an external download manager

ArjixGamer:
wget cant create directories that we specify, it can only create its own url scraped directories which is not desired, so for now it will only download to the Current Working Directory. While the download path doesnt work as desired, we can pass the referer to make it work with providers that require a referer, any ideas on how to get the download dir working are welcome


Blatzar:
I thought anime dl made the directories needed, I've at least seen some function in utils to make directory or something similar.

ArjixGamer:
> I thought anime dl made the directories needed, I've at least seen some function in utils to make directory or something similar.

it does to an extent, but we also make a directory with file_format

Blatzar:
Doesn't seem to work on twist :/

AbdullahM0hamed:
UPDATED: now works on twist, and download path is created prior to downloading with wget

IguanasInPyjamas (commented):
not relevant

AbdullahM0hamed:
@IguanasInPyjamas this doesn't touch idm in any way. Wget has Linux supports, and I've tested that it works on android

IguanasInPyjamas:
> @IguanasInPyjamas this doesn't touch idm in any way. Wget has Linux support, and I've tested that it works on android

OOPS, yeah I got confused in my head. Lemmie delete that

IguanasInPyjamas:
I'll test wget on Linux, can someone test wget on windows? then I'll move to merge

nate-moo:
got it for windows

nate-moo:
it doesn't work

nate-moo:
image
works

IguanasInPyjamas:
Works on Linux and twist. Moving to merge

PR #461, opened 2020-08-01
Re-implementing watch 

MichaelDavidJohnson:
It seems that the overall cli is nicely formatted, it's mainly a case of removing spaghetti and updating the code base while allowing for future developments. Leaving this as a draft while I work on the rest

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,7 @@ Yeah. Me too! That's why this tool exists.
 ## Features
 
 - Download or stream any episode or episode range of any anime.
+- Have a locally stored anime list to track your progress and stream anime using the watch command.
 - Specify the quality you want to stream or download.
 - Search and download.
 - Save yourselves from those malicious ads.
--- a/anime_downloader/commands/watch.py
+++ b/anime_downloader/commands/watch.py
@@ -130,7 +130,7 @@ def list_animes(watcher, quality, download_dir,imp=None):
             meta += '{}: {}\n'.format(k, click.style(str(v), bold=True))
         click.echo(meta)
 
-        click.echo('Available Commands: set, remove, update, watch,'
+        click.echo('Available Commands: set, remove, update, watch, back'
                    ' download.\n')
 
         inp = click.prompt('Press q to exit', default='q').strip()
@@ -139,6 +139,8 @@ def list_animes(watcher, quality, download_dir,imp=None):
         # Decorator?
         if inp == 'q':
             break
+        elif inp == 'back':
+            list_animes(watcher, quality, download_dir, imp=None)
         elif inp == 'remove':
             watcher.remove(anime)
             break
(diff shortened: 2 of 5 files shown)


MichaelDavidJohnson:
Okay so, current developments and errors 

- Providers have been updated and will autoupdate the same way as dl
- Play actually works, hardcoded with mpv right now, going to add an option for adding and changing that 
- Seem to get a non-type error after it loads one episode, everything still functions but that needs to be fixed. I think all the non-type is doing is kicking us out of the watch environment 

If anyone has any optimisation ideas about removing some of the spaghetti in the code base send them my way, I'm focusing on merely getting it running and usable for users at the moment. A refactor is hard when the original code doesn't even work.

Blatzar:
Will probably take a good look at it tomorrow.  A big refractor isn't out of the question. Haven't used watch so you'll have to fill me in on the details.

I think watch uses the player folder, but since it's already done in utils I think it's better to use that already finished implementation.

MichaelDavidJohnson:
Yeah I'm starting to drag in utils

MichaelDavidJohnson:
Okay so now
- When you watch a show it will always bring you back to that show, to watch again.

Things to be added 
- A way to re-enter the list from a show without using CTRL-C
- A way to queue multiple episodes in watch (it may already be there, the pasta sauce is covering it )
- Improve the docs / tutorial for the use case of watch 
- Local serialisation of the list into a MAL like format so we can then import and export MAL lists.

MichaelDavidJohnson:
For context, this PR is just to regain functionality of watch using the old codebase, it literally wasn't working before. 

A rewrite will happen once the watch command is functional again

Blatzar:
Download still seems to use the basedownloader instead of aria2
Watch loops the same episode forever until exit.

MichaelDavidJohnson:
Yeah downloader hasn't been touched yet

Watch shouldn't loop the episode, it should return back to the play menu

MichaelDavidJohnson:
Fixed the weird looping issue. Next it to set up downloader to choose between aria and default, lagrad should be working on this.

Blatzar:
It now uses the same downloading system as dl does, including the same config

MichaelDavidJohnson:
Very nice, we are ready to merge to fix the spaghetti! And then it's time to refactor it all

Blatzar:
Bug: Using back multiple times will make q (quit) not work as intended. If you use back 10 times you'll have to spam q 10 times to actually exit the script.

Blatzar:
Bug (ish): you can set episodes_done above title length (also causing some funky behavior).

MichaelDavidJohnson:
What the heck tests are you doing to find these bugs :o

Blatzar:
Bug: Assuming valid input on set episodes_done causing errors if not int. We'd want at least graceful exit or reprompt.

Blatzar:
Bug: invalid provider name errors watch, graceful exit or reprompt would be better

MichaelDavidJohnson:
dude the teamwork on this PR is wild, 100% efficiency.

Blatzar:
Bug: inconsistent download, anime dl -e 1:3 downloads ep 1-3, while watch downloads 1-2

Issue #257, opened 2019-10-25
Improve security of javascript execution


LiteracyFanatic:
I'm no expert on NodeJs or security in general, but I wanted to start a discussion about `eval_in_node`. At the moment we are executing what is essetially arbitrary js in two different locations in the codebase. While it may be known what the scripts that are being fetched and executed do now, if they were to change in the future or the sites were to be compromised in some way, significant damage could be done since node is able to do pretty much anything the calling user is capable of.

Probably the only truely safe way to do this is using a container, but that would be a lot of overhead to fetch a cookie. Again I'll stress that I'm not sure what the proper solution would be, but perhaps [vm2](https://github.com/patriksimek/vm2) would be a decent approach. vm2 blocks node's modules by default and allows whitelisting needed functionality. It looks like it would just take a small wrapper script to lock down the node process quite a bit. That way could still log the cookies but forbid the process from starting other processes or accessing the file system.

Any thoughts about this or other possible approaches to the issue? The only downside I see with vm2 would be the addition of another dependency to the project. If this seems like a good addittion to you, I can write up a pull request.

MichaelDavidJohnson:
This is quite a major problem actually, you are completely right that it is a security issue. Can you please come back and work on this ;-;

PR #463, opened 2020-08-02
Sandbox execution of JavaScript with vm2

LiteracyFanatic:
Addresses #257.

It's been a while since I've looked at this project, but as far as I can tell everything still works with these changes. I'm not terribly familiar with python packaging. What would be the proper way to incorporate the `node_modules/` into the project's dependencies?


MichaelDavidJohnson:
That was very quick, thank you very much. I'll put 0x up to review this one, as it's quite major.

LiteracyFanatic:
No problem, I needed something similar for one of my own projects awhile back. The sandboxed code doesn't have access to `require()` or any of the builtin objects `node` normally providers like `process`. So it can evaluate JavaScript expressions, but it can't access the file system or start other processes. It is possible to pass a whitelist of modules to allow if their functionality is needed.

lord8266:
You should also encode the js to base64, otherwise one can run code in the global context like so
```py
eval_in_node('`+console.log(process.env)+`')
```
Something like this should do
```py
def eval_in_node(js: str):
    js = base64.b64encode(js.encode('utf-8')).decode()
    sandboxedScript ="""
                    const {VM} = require('vm2'); 
                    const js = Buffer.from('%s','base64').toString()
                    console.log(new VM().run(js))
                    """%js
    output = subprocess.check_output(['node', '-e', sandboxedScript])
    return output.decode('utf-8')
```

LiteracyFanatic:
Good catch @lord8266. I've just pushed an updating incorporating your changes.

vn-ki:
This is something I have thought of too, but I didn't see a very straightforward way to solve it.

How do you propose we install the vm2 dependency?

LiteracyFanatic:
Alright, I did some reading and it looks like `setuptools` will only let us include files located under the package directory. So if we move `node_modules/` under `anime_downloader/` and add `package_data={ 'anime_downloader': ['node_modules/**/*'] }` to `setup.py` it should get included. As far as making sure `node_modules` is present, I see three options.

1. Commit `node_modules/` to source control. This is generally considered a bad practice, but it's only a single package with no external dependencies so it's not too much of a problem here.
2. Only commit `package.json` and `package-lock.json` to source control. This is cleaner, but we'll have to always remember to run `npm install` when cloning or building the package.
3. Add some scripting to `setup.py` to handle the `npm` stuff automatically.

Thoughts?

Blatzar:
> Thoughts?

Npm stuff goes over my head, but I'd much prefer any option to make it as automatic as possible. I'd rather not add another step for people using cli for their first time in their life.

LiteracyFanatic:
End users wouldn't be affected in any of those scenarios assuming they are using they're just installing the package with `pip` or their operating system's package manager.` The issue is about whether we vendor the npm dependencies (make them a part of the git repo) or have developers install them when cloning the branch. If you don't have strong feelings about this, I'll just commit `node_modules/` for now since that's the simplest approach.

LiteracyFanatic:
I just pushed an update with the `node_modules/` included in the package.

vn-ki:
Is the node modules present inside the installation directory automatically picked up by node?

LiteracyFanatic:
Ah, that's right. The executable ends up in a different folder. We can fix that by setting the environment variable `NODE_PATH` for the subprocess. I'm not sure how to figure out the package installation directory programatically though.

LiteracyFanatic:
Figured it out. It should find the folder now.

vn-ki:
Alright! This looks good now! Thanks for the contribution.

We can merge this after some testing. (I want someone to test this on windows)

Blatzar:
@vn-ki Just testing twist.moe (the one who seems most effected looking at diff) it works without any issues on windows 10.

PR #467, opened 2020-08-10
aria2 default external downloader.

Blatzar:
When aria2 isn't installed it will prompt the user to install it, but when it's installed it won't use it since the previous config generated using no external downloader will be used. This results in a confusing install, especially with aria2 being in curly brackets.

TODO: check what happends in a clean install without aria2 when config uses aria2

### DIFF
--- a/anime_downloader/config.py
+++ b/anime_downloader/config.py
@@ -18,7 +18,7 @@ DEFAULT_CONFIG = {
         'force_download': False,
         'file_format': '{anime_title}/{anime_title}_{ep_no}',
         'provider': 'twist.moe',
-        'external_downloader': '',
+        'external_downloader': '{aria2}',
         'aria2c_for_torrents': False,
         'selescrape_browser': None,
         'selescrape_browser_executable_path' : None,


MichaelDavidJohnson:
Second to remove curly brackets from external downloaders, I never understood why it was like that in the first place and I always forget them :P.. That being said, if you're proposing aria2 to be the new default what are your intentions with the internal downloader?

Blatzar:
Since it errors on no aria2 in path it's pretty much aria2 default already, I'd love to have it use an internal downloader as default if aria2 wasn't so much better. I don't even think the internal one works in twist.

I def agree on the curly brackets, I see no logical reason to make that config special.

Blatzar:
Since it errors on no aria2 in path it's pretty much aria2 default already, I'd love to have it use an internal downloader as default if aria2 wasn't so much better. I don't even think the internal one works in twist.

I def agree on the curly brackets, I see no logical reason to make that config special.

MichaelDavidJohnson:
Yeah, I definitely personally think that the internal downloader either needs some rework to try to gain performance (the gap is just too big) or just needs to be removed in favour of potentially adding more user choice to the external downloaders.. Just my opinion though.

Blatzar:
Deleting curly brackets would be easy but would break for users upgrading. Nothing we can't sidestep, but will result in italian cuisine :)

vn-ki:
Won't this cause weird ass errors when aria2 isn't installed?

Blatzar:
> Won't this cause weird ass errors when aria2 isn't installed?

When aria2 isn't installed this already happens:
image

vn-ki:
I have plans to revive internal downloader, using parfive. If you have time, you can investigate whether using parfive is worth it.

Blatzar:
> I have plans to revive internal downloader, using parfive. If you have time, you can investigate whether using parfive is worth it.

I can't get parfive to download from twist at least. Might've missed something.
image

vn-ki:
Alright, let's merge this for now.

Blatzar:
Closing as pySmartDl is a good replacement

PR #470, opened 2020-08-15
Kiss anime is dead, remove from init, possibly remove site file aswell

nate-moo:


Blatzar:
Wont remove the site file in case copies gets made

nate-moo:
> Wont remove the site file in case copies gets made

ok

PR #487, opened 2020-08-22
Added a custom extractor that can be used if a provider wants to add the referer by itself.

ArjixGamer:
It will split the url using a newline (\n), the 1st index will be the stream_url and the 2nd index will be the referer.


vn-ki:
I don't like this at all.

The ideal way be to add refer in the constructor of `BaseExtractor` and import it in the provider which needs to specify the referer.

Issue #486, opened 2020-08-22
Js sandbox causing errors


Blatzar:
**Describe the bug**

https://github.com/vn-ki/anime-downloader/commit/9e254ccd2b85eb72cd40ecef36a4b937c13e3dfb seems to have caused issues with eval_in_node()

Without modifications it gives `NameError: name 'path' is not defined`

Editing https://github.com/vn-ki/anime-downloader/blob/master/anime_downloader/util.py#L402 to os.path will result in error below

**To reproduce**

```
anime -ll DEBUG dl "hunter x hunter" --provider animekisa -c 2
```

<details>
<summary>LOG</summary>

```
bl@arch ~> anime -ll DEBUG dl "hunter x hunter" --provider animekisa -c 2
2020-08-22 14:01:28 arch anime_downloader.util[84260] INFO anime-downloader 4.6.3
2020-08-22 14:01:28 arch anime_downloader.util[84260] DEBUG Platform: Linux-5.7.7-arch1-1-x86_64-with-glibc2.2.5
2020-08-22 14:01:28 arch anime_downloader.util[84260] DEBUG Python 3.8.3
2020-08-22 14:01:28 arch anime_downloader.sites.helpers.request[84260] DEBUG -----
2020-08-22 14:01:28 arch anime_downloader.sites.helpers.request[84260] DEBUG GET https://animekisa.tv/search
2020-08-22 14:01:28 arch anime_downloader.sites.helpers.request[84260] DEBUG {'params': {'q': 'hunter x hunter'}}
2020-08-22 14:01:28 arch anime_downloader.sites.helpers.request[84260] DEBUG {'user-agent': 'Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/536.5 (KHTML, like Gecko) Chrome/19.0.1084.9 Safari/536.5'}                                        
2020-08-22 14:01:28 arch anime_downloader.sites.helpers.request[84260] DEBUG -----
2020-08-22 14:01:28 arch anime_downloader.session[84260] DEBUG cached request
2020-08-22 14:01:28 arch anime_downloader.sites.helpers.request[84260] DEBUG https://animekisa.tv/search?q=hunter+x+hunter                                                                                                                  
2020-08-22 14:01:28 arch anime_downloader.sites.helpers.request[84260] DEBUG /tmp/animedlm6tlqktt/tmpzys605nr
+--------+--------------------------------------------------+--------+
[...]

PR #491, opened 2020-08-23
Fix eval

LiteracyFanatic:
This addresses #486. It will almost definitely have broken other places that use `eval_in_node`. I'll add commits fixing those later. I'll also write some guidelines on how to use the js evaluation since it's gotten a bit more complicated than before. Hopefully the end result will be more robust and rely less on parsing scripts as strings.

### DIFF
--- a/anime_downloader/util.py
+++ b/anime_downloader/util.py
@@ -6,6 +6,7 @@ import subprocess
 import platform
 import re
 import os
+from os import path
 import json
 import errno
 import time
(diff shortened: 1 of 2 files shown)


vn-ki:
A fallback (with a warning) if it fails would be good.

anime_downloader/util.py
```diff
@@ -390,16 +391,20 @@ def get_hcaptcha_cookies(url):
         return pickle.load(open(COOKIE_FILE, 'rb'))
 
 def deobfuscate_packed_js(packedjs):
-    return eval_in_node('eval=console.log; ' + packedjs)
+    return eval_in_node(packedjs)
 
-def eval_in_node(js: str):
+def eval_in_node(js: str, sandbox: str = "{}"):
```
vn-ki, line 396: Shouldn't this be transparent? We can't assume the next developer (who's going to use `eval_in_node` for implementing a provider) to know how to construct a sandbox string.

ArjixWasTaken:
We should use https://github.com/amol-/dukpy instead, that would also remove our dependency for nodejs.

PR #492, opened 2020-08-24
changed how the episodes list is handled for anime8

ArjixGamer:
even if eps.pop() was not faulty, its not bad to remove it just in case, i added a list and a for loop for all the eps that we may want to skip.
suggestion: adding those special eps at the end of the list instead of removing them, because some anime are only made of special eps.


ArjixGamer:
not for merge, i dont remember how, but i made a huge mistake

Issue #493, opened 2020-08-25
4anime download problem


LordShenron:
**Describe the bug**

4anime recently moved servers as per their notice and because of this URLs returns 403. 

**To reproduce**

I tried with Black clover's episode 140. Downloader returns a 403 not found for the episode but When i tried to see if the episode is actually there or not on the website, it was there. And I was able to download the file. 

```
anime --log-level DEBUG dl 'black clover' --provider 4anime --episodes 140
```

<details>
<summary>LOG</summary>

```
┌─[shen][wolf][~/Downloads/Video/Black-Clover]
└─▪ anime --log-level DEBUG dl 'black clover' --provider 4anime --episodes 140
2020-08-25 21:22:09 wolf anime_downloader.util[17291] INFO anime-downloader 4.6.3
2020-08-25 21:22:09 wolf anime_downloader.util[17291] DEBUG Platform: Linux-4.19.0-10-amd64-x86_64-with-MX-19.2-patito_feo
2020-08-25 21:22:09 wolf anime_downloader.util[17291] DEBUG Python 3.7.3
2020-08-25 21:22:09 wolf anime_downloader.sites.helpers.request[17291] DEBUG -----
2020-08-25 21:22:09 wolf anime_downloader.sites.helpers.request[17291] DEBUG POST https://4anime.to/wp-admin/admin-ajax.php
2020-08-25 21:22:09 wolf anime_downloader.sites.helpers.request[17291] DEBUG {'data': {'action': 'ajaxsearchlite_search', 'aslp': 'black clover', 'asid': 1, 'options': 'qtranslate_lang=0&set_intitle=None&customset%5B%5D=anime'}}
2020-08-25 21:22:09 wolf anime_downloader.sites.helpers.request[17291] DEBUG {'user-agent': 'Mozilla/5.0 (Windows NT 6.2; WOW64) AppleWebKit/535.11 (KHTML, like Gecko) Chrome/17.0.963.66 Safari/535.11'}
2020-08-25 21:22:09 wolf anime_downloader.sites.helpers.request[17291] DEBUG -----
[...]

PR #494, opened 2020-08-26
Allowing for headers to be passed to downloader and fixed 4anime

Blatzar:
Fixes https://github.com/vn-ki/anime-downloader/issues/493 using the internal downloader. 
Aria2 doesn't seem to work with 4anime though.


IguanasInPyjamas (commented):
I'm happy with this.

PR #501, opened 2020-08-30
added a docstring, why? /shrug

ArjixGamer:


ArjixGamer:
in this pr ill comment my providers

PR #503, opened 2020-09-01
added a cache system for SeleScrape and Improve SeleScrape generally

ArjixWasTaken:
Now all requests are cached for 1 hour, then they become invalid.
Hmm, this is draft until i fix a bug i just found.


ArjixGamer:
Its working as it is, ill add some comments and docstrings and then its ready for merge

ArjixGamer:
marked as draft because i will change the cache expiration logic

ArjixGamer:
i need help, with the unix timestamps i cant get it to ignore the invalid cached request

ArjixGamer:
oh wait, it works, im just bad at math

ArjixGamer:
im confident that its ready for review

ArjixGamer:
making this a draft once again because ill rewrite and rearrange a lot of the spaghetti code (0x said it is, so ill believe him).

IguanasInPyjamas:
Isn't this big broken?

AbdullahM0hamed (changes requested):
The bool cast here seems to be unnecessary.

anime_downloader/sites/helpers/selescrape.py
```diff
@@ -85,113 +128,115 @@ def driver_select(): #
...
-    status = status_select(driver, new_url, 'hide')
-    try:
-        cloudflare_wait(driver)
-        user_agent = driver.execute_script("return navigator.userAgent;") #dirty, but allows for all sorts of things above
-        cookies = driver.get_cookies()
-        text = driver.page_source
-        driver.close()
+    url = url if not params else url + '?' + urlencode(params)
+
+    if bool(check_cache(url)):
```
AbdullahM0hamed, line 209: Bool cast here is unnecessary

anime_downloader/sites/helpers/selescrape.py
```diff
@@ -85,113 +128,115 @@ def driver_select(): #
...
+        cached_data = check_cache(url)
```
AbdullahM0hamed, line 210: Instead of running the method twice on the same url, perhaps store in a variable and check the variable?

AbdullahM0hamed (changes requested):

anime_downloader/sites/helpers/selescrape.py
ArjixGamer: I like to play it safe, so it's not unnecessary. It also doesn't impact the code in any way, so is there a need to remove it?

ArjixWasTaken:
### If anything is left to improve that would be 2 things:
1. Add the ability to use selenium from the provider directly.
2. Improve the cloudflare bypass (of just waiting it out)

ArjixWasTaken:
oops, found an error

ArjixWasTaken:
> oops, found an error

turns out my config file was just corrupted

ArjixWasTaken:
Ok, I have now tested on both firefox and chrome (on windows).
If anybody is willing to test on linux I would be happy.

You can run this code to test
```python
import time

f = helpers.get('https://gogoanime.vc/', sel=True, cache=False)  # will not save cache
print("Successfully ran w/o saving cache!")


f = helpers.get('https://gogoanime.vc/', sel=True)  # will save cache
print("Successfully ran and saved cache!")

start = time.time()

f = helpers.get('https://gogoanime.vc/', sel=True)  # will read cache
end = time.time()

if end-start < 5:
    print("Successfully used cache!")
```

after that wait 30 minutes, and then run

```python
import time

start = time.time()
f = helpers.get('https://gogoanime.vc/', sel=True)  # will read cache
end = time.time()

if end-start > 5:
    print("Cache was expired and retrieved new cache successfully!")
else:
    print("Cache has not expired yet.")
```

AbdullahM0hamed (approved):

ArjixWasTaken:
Merged this, but alas, no provider uses selescrape anymore.

PR #505, opened 2020-09-06
added topmovies

Blatzar:
Leaving this here until m3u8 downloader gets implemented as it seems like all their files are m3u8.


AbdullahM0hamed:
```505 HTTP Version Not Supported```

Blatzar:
Sloppily merged with m3u8 downloader for testing, will fix later when m3u8 is finished

IguanasInPyjamas:
Left.

Issue #507, opened 2020-09-07
Something is up with Aria2


retardedloozer:
Trying to download episodes using aria2 seems to yield alot of errors these days. Trying to download Magia Record from twist.moe yields me the error 403 message yet when I managed to remove Aria2 from the config.json, allowing me to use the slow internal downloader, I'm able to download anime episodes, albeit at a reduced rate. 

I think this issue describes whats going on perfectly -> https://github.com/vn-ki/anime-downloader/issues/493

Blatzar:
Known issue and the reason for https://github.com/vn-ki/anime-downloader/pull/498 however the reason for #498 not being merged yet is because of https://github.com/vn-ki/anime-downloader/pull/504, we have yet to decide on which the replacement is going to be. 

As for aria2, I have yet to find a good reason for the sudden 403 errors across providers and devices. My best guess is that cloudflare pushed some new tech which detects aria2. Editing command line arguments doesn't seem to help hence I think it's some core issue unfixable by us.

retardedloozer:
> Known issue and the reason for #498 however the reason for #498 not being merged yet is because of #504, we have yet to decide on which the replacement is going to be.
> 
> As for aria2, I have yet to find a good reason for the sudden 403 errors across providers and devices. My best guess is that cloudflare pushed some new tech which detects aria2. Editing command line arguments doesn't seem to help hence I think it's some core issue unfixable by us.

Roger that. Thank you for your hard work anyway. I'll close this

PR #510, opened 2020-09-13
Unquote user-agent for aria2c

AbdullahM0hamed:


AbdullahM0hamed:
Should probably mention #507

AbdullahM0hamed:
```510 Not Extended```

PR #513, opened 2020-09-15
Added provider Tenshi.moe

ArjixGamer:
https://tenshi.moe/


ArjixGamer:
I think I can't improve it anymore than this, well... I guess I can replace the lambda with a CSS selector to find by partial attribute value...but I can't find the Stack Overflow question that has that method.

Blatzar:
Add to Readme on both PR:s

ArjixGamer:
> Add to Readme on both PR:s

done

PR #517, opened 2020-09-16
fixed AnimeFrenzy

ArjixGamer:
The website has changed a lot since the creation of the provider, I use the old website for the scraping so I don't know how long this fix will last, but since its the old design it means it wont get any changes. Only change I can think of is the deletion of the old webpage portal.

### DIFF
--- a/anime_downloader/config.py
+++ b/anime_downloader/config.py
@@ -46,7 +46,8 @@ DEFAULT_CONFIG = {
     },
     'siteconfig': {
         'animefrenzy': {
-            'version': 'subbed'
+            'version': 'subbed',
+            'servers': ['trollvideo', 'mp4upload']
         },
         'egyanime': {
             'version': 'subbed',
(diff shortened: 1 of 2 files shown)


ArjixGamer:
forgot to remove logger object after removing the import, fixed it.

ArjixGamer:
added servers list in config

Blatzar:
This is the last mp4upload/trollvid provider I'm merging to this repo, we already have plenty of those copies. Also you're missing xstreamcdn from config

IguanasInPyjamas:
Some conflicts need to be resolved.

ArjixGamer:
making this pr a draft, bc im not really sure animefrenzy will stay as it is for long

ArjixGamer:
this might be useful at a later point in time

AbdullahM0hamed:
#630

ArjixGamer:
this PR still works (kinda) it only has the old servers of animefrenzy...(the new animefrenzy only has one server but it always works)
image

AbdullahM0hamed:
Want to replace the current implementation? Or implement the old version as a fallback?

ArjixGamer:
I'll implement the old version as a fallback I guess

ArjixGamer:
woah, i completely forgot about this pr

Issue #489, opened 2020-08-22
Skip vidstream m3u8


Blatzar:
**Describe the feature here**
Currently vidstream allows downloading m3u8 files and names them .mp4 which confuses users, I'd like to see either proper m3u8 extension for those files or vidstream skipping them (through config) or both.

vn-ki:
Just implement a m3u8 downloader xD

IguanasInPyjamas:
> Just implement a m3u8 downloader xD

cries in FFMPEG

gingerbeardman:
`ffmpeg -i http://.../playlist.m3u8 -c copy -bsf:a aac_adtstoasc output.mp4`

from: https://stackoverflow.com/a/32537953/28290

Blatzar:
Ffmpeg is too big of a dependency to get accepted into this project, hence why we haven't done that already.

ArjixGamer:
Can't we at least add support for it?
It won't be a dependency but we can add support for it if a user is willing to download it / has already downloaded it.

gingerbeardman:
Exactly my thoughts.

IguanasInPyjamas:
By that respect, you can simply use FFMPEG on the file from the command line yourself. That's where the support is, you already HAVE the file.

The only option that I think is necessary is a small m3u8 downloader,  there's no point adding a huge dependency like FFMPEG when you can just choose the alter the files with that yourself.

gingerbeardman:
Good point.

I guess the issue is the file needs to be renamed to m3u8 and the user prompted that it's ready for conversion, perhaps the ffmpeg command line could even be quoted in the output?

IguanasInPyjamas:
100% m3u8 files should be .m3u8 if we don't convert ourselves.

PR #519, opened 2020-09-18
M3u8 support 

Blatzar:
Not tested on windows yet, fixes https://github.com/vn-ki/anime-downloader/issues/489


IguanasInPyjamas (dismissed):
I have no issues with this.

Blatzar:
All the fuckery because external downloader and internal downloader is treated very diffrently

Blatzar:
Doesn't support encryption or this shit below, I'll look into them both.
I think I got encryption working, but it still wont work fully on animepahe, don't know if it's because they fuck their videos or because the decryption is lacking.
```
#EXTM3U
#EXT-X-VERSION:3
#EXT-X-STREAM-INF:BANDWIDTH=800000,RESOLUTION=640x360
360p/360p.m3u8
#EXT-X-STREAM-INF:BANDWIDTH=2800000,RESOLUTION=1280x720
720p/720p.m3u8
```

IguanasInPyjamas:
Issue with m3u8s from searches of the form ``anime dl -u "code geass" -p animekisa -c 5 `` 

The issue comes from 'segments'. This returns an empty array inside ``self.m3u8file.get_tssegments()``, which leads to an M3u8DownloaderNoStreamException().

PR #520, opened 2020-09-18
Remove gogoanime

Blatzar:
Gogoanime provides the same sources as vidstream, gogoanime works, but to ease maintenance I'm removing duplicate providers.


IguanasInPyjamas:
Uhhh, I don't mind it being removed. It's basically normie handling for people who only know websites and not servers

PR #521, opened 2020-09-19
Change how config is read

AbdullahM0hamed:
This changes the method of reading from config.json from `json.load` to `eval`.

Why? Because `json.load` doesn't allow backslashes in config.json (unless they are themselves escaped) - which is a problem to windows users.

Tested on my config. Tested on config after deleting config. 

Could stand to be tested further by other people.

### DIFF
--- a/anime_downloader/config.py
+++ b/anime_downloader/config.py
@@ -232,7 +232,7 @@ class _Config:
     def _read_config(self):
         with open(self.CONFIG_FILE, 'r') as configfile:
             try:
-                conf = json.load(configfile)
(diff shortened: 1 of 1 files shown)


anime_downloader/config.py
```diff
@@ -232,7 +232,7 @@ def _write_config(self, config_dict):
     def _read_config(self):
         with open(self.CONFIG_FILE, 'r') as configfile:
             try:
-                conf = json.load(configfile)
+                conf = eval(configfile.read().replace("true", "True").replace("false", "False").replace("null", "None"))
```
vn-ki, line 235: `eval`ing a foreign file is bad practice.

PR #525, opened 2020-09-21
Remove unnecessary download_episodes function

AbdullahM0hamed:
Not really used anywhere: https://github.com/vn-ki/anime-downloader/search?q=download_episode&type=

Function can easily be replicated by doing `episode.download()`.

`dl`, `ezdl`, and `download` (using `watch`) have been tested to work with both aria2c and internal (barring `download` from `watch` which fails before the changes and after the changes as well when used with internal - works fine on aria2c)

### DIFF
--- a/anime_downloader/util.py
+++ b/anime_downloader/util.py
@@ -207,6 +207,7 @@ def print_episodeurl(episode):
     # Currently I don't know of a way to specify referer in url itself so leaving it here.
     print(episode.source().stream_url)
 
+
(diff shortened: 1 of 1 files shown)


Blatzar:
Merging soon™, do you think it's ok to remove bypass_hcaptcha() too as it's outdated and probably won't be worked on?

AbdullahM0hamed:
Yeah, should be removed

PR #526, opened 2020-09-26
PyQt GUI

IguanasInPyjamas:
So, Originally I wanted this front end to be a placeholder while Soldat improves my previous UI. However, I really like the clean and simple look of this GUI. It seems to suit the aesthetic of anime-dl in my eyes. So, I'm probably going to pursue this GUI for the foreseeable future. 

Only one extra dependency - PyQt5. You're gonna need a dependency no matter what we do if we want to not build a GUI environment from scratch. 

I've added all the "normie" features, episode slices still works, Nyaa still works even with the episodes options.

Ezdl has not been included yet, I'm going to implement a tick box for ezdl. However, is it even really necessary in this GUI.

Theming options will be coming soon, devs and collabs will have the option to give a specific theme for them.

I personally feel like a GUI like this is superior to the cloudfare port alternative. Firstly, the code is simple enough for other people to maintain if I ever leave. Secondly, the *Netflix style* doesn't really suit anime-dl especially when we don't have a native player. Another advantage is the use of actual anime-dl commands. 

I haven't added a GUI command to the anime cli, I would suggest people just adding the script to their hotbar etc. as that allows the normies to just use a GUI.

This PR is a draft, mainly because I need to add comments especially around the threading component, take feedback on theming options from other devs and feature additions.

Any feedback would be appreciated amigos.


ArjixGamer:
add PyQt5 as a requirement in setup.py

IguanasInPyjamas:
Good catch.

IguanasInPyjamas:
Issue -> PySmartDL breaks the internal downloader, was fine before I updated to that... something is going wrong with the backwards compatibility.

Blatzar:
PySmartDL replaced the old internal downloader, how are you accesing the internal?

ArjixGamer:
image

Blatzar:
It cannot be accessed in that way, using episode.download() if there's no external is how it's done.

image

Blatzar:
Feedback:
"..." to select download dir isn't very intuitive imo, might be placeholder.
A play button would be nice. I could probably contribute with those though.

anime_downloader/config.py
```diff
@@ -9,7 +9,7 @@
 DEFAULT_CONFIG = {
     'dl': {
         'url': False,
-        'player': None,
+        'player': 'mpv',
```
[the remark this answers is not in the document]
Blatzar: https://github.com/IguanasInPyjamas/anime-downloader/pull/5
for fix

IguanasInPyjamas:
Issue -> Terminating the thread when a download is activating either doesn't stop the download or full errors out the app. We need some handling for this

Blatzar:
Features needed:
Default theme in config
Default provider in config

Bug: Spamming the download button will start multiple downloads and eventually kill the gui while the threads are running in the background

AbdullahM0hamed:
> we only have like 4 themes

Probably better to add theme settings in which the user can select the colours of certain aspects of the UI as opposed to adding more

IguanasInPyjamas:
@AbdullahM0hamed feel free to autoPEP8 this for me :)

setup.py
```diff
@@ -30,6 +30,7 @@
         'cfscrape>=2.0.5',
         'requests-cache>=0.4.13',
         'tabulate>=0.8.3',
+        'PyQt5>=5.15.1',
```
AbdullahM0hamed, line 33: Pyqt doesn't install on android (at least via termux) - so this breaks anime-downloader on android, and it's only needed by people who will use the GUI - perhaps move this to `extras_require`?

AbdullahM0hamed (changes requested):

setup.py
IguanasInPyjamas: That's a great idea. I'll do that

AbdullahM0hamed (approved):

PR #527, opened 2020-09-30
Added cache flag to helpers

Blatzar:
Can be used like `req = helpers.get(self.url, referer=referer, cache=False)`
Prevents reading from cache, useful for tokens unique to each session


vn-ki:
You shouldn't create a new session every time. This will mean that the cookies from the consecutive requests will not be persisted.

Other than that, I like the change.

Blatzar:
I see no feasible way of doing persistent cookies without hacks, also I'm pretty sure only storing the cookies is bound to mindfuck people later on using the feature.

> You shouldn't create a new session every time

Having a separate session cache with an expire_after of something like 10s is possible, would that be good?

vn-ki:
> Having a separate session cache with an expire_after of something like 10s is possible, would that be good?

That's what i had in mind, yes

> I see no feasible way of doing persistent cookies without hacks, also I'm pretty sure only storing the cookies is bound to mindfuck people later on using the feature.

I did not mean persistent cookies across runs. Just in a single run, the cookies shouldnt be invalidated. If you are creating a new session every time, then the cookies are not persisted.

Blatzar:
> That's what i had in mind, yes

Fixed in latest commit, should I rename the flag to something more appropriate?

AbdullahM0hamed:
Arigato for your work

PR #530, opened 2020-10-01
Redid slugify regex for better titles

Blatzar:
Will now allow +-! and disallow the file starting with .
Didn't add ' and " to the whitelist as it's not as console friendly.

### DIFF
--- a/anime_downloader/util.py
+++ b/anime_downloader/util.py
@@ -242,12 +242,13 @@ def get_json(url, params=None):
 
 def slugify(file_name):
     file_name = str(file_name).strip().replace(' ', '_')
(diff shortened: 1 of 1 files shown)


ArjixGamer:
em, unless you specifically input ' and " in the code itself, that wont be a problem, python escapes those characters

Blatzar:
> em, unless you specifically input ' and " in the code itself, that wont be a problem, python escapes those characters

I mean in terminal shell

ArjixGamer:
image

ArjixGamer:
oh wait, echo is not the best example

Blatzar:
I'd like to avoid backslashes in my filenames, a consequence of ' and "
image

PR #531, opened 2020-10-04
Update anime.py change "QUALITIES" to a tuple

Sultron:


AbdullahM0hamed:
Not sure that the difference will even be noticeable, but you might as well comb through the providers and change there as well, e.g:

https://github.com/vn-ki/anime-downloader/blob/f13f13bc367c7df8f6d667d7ff74266b673386a2/anime_downloader/sites/animefreak.py#L12

Blatzar:
@Sultron please merge all PR:s to one, also I'm not even sure QUALITY in providers matter much, if at all.

IguanasInPyjamas:
There's nothing particularly wrong with using tuples instead of lists, I doubt the memory and time save will be too significant with this change.

IguanasInPyjamas:
No change on this PR after asking an action of all those PRs to be merged into here. Reopen when you do that.

PR #542, opened 2020-10-05
add putlockers

AbdullahM0hamed:
Works for internal, but not aria2c atm. Will look at that tomorrow - gonna open this as draft for now


AbdullahM0hamed:
Changed default user agent, will not to be tested to see how it affects other providers

PR #550, opened 2020-10-09
Restore circleci config.yml

cxong:
This is just a quick build and test with circleci, so that the commits are green again. Not sure what the plan is though, if we want to have circleci do the exact same builds as travis, or if circleci should be removed altogether.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,49 +1,18 @@
-# pip-run: &pip-install
-#   name: Install Python dependencies
-#   command: |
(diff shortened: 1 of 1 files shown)


vn-ki (approved):
For now, this is fine I guess. IIRC Travis is gonna say goodbye soon, so this would be a backup.

cxong:
Are there any changes to be made before this can be merged?

IguanasInPyjamas:
Oh, it was already approved. Merging

PR #560, opened 2020-10-18
Animefreak: Set Episode Number manually when possible

AbdullahM0hamed:
Sometimes the episode count doesn't match the number of episodes (maybe due to missing episodes?). An example of this is One Piece, where the latest episode, 946, downloads as `Episode 942` instead of `Episode 946`

### DIFF
--- a/anime_downloader/sites/animefreak.py
+++ b/anime_downloader/sites/animefreak.py
@@ -45,7 +45,7 @@ class AnimeFreakEpisode(AnimeEpisode, sitename='animefreak'):
         # E.g. episode-946
         url_end = self.url.split("/")[-1]
 
-        # To ensure that isn't a special/preview or something
+        # To ensure that it isn't a special/preview or something
         logger.info(url_end)
         if "episode" in url_end:
             episode_number_regexed = re.search("\d+", url_end)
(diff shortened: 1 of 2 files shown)


AbdullahM0hamed:
Actually, should probably do something about selection before merging this, since -e 946 (as per the example) would still not work

IguanasInPyjamas:
Yeah, the naming convention wouldn't change the fact it's only getting 942 instead of 946 because all the specials are front loaded or whatever on the provider

AbdullahM0hamed:
The naming convention change itself is still nice, because after having watched up to "942", I found out from livechart.me that there are 946 eps arm, and I was very confused lol

vn-ki (commented):
I think when you are returning the links (from `scrape_episodes`) you can return a list of tuples `[(ep_no, ep_link)]` instead. (I remember writing logic for this, not sure if it still exists. It's worth checking out.)

AbdullahM0hamed:
Definitely a step in the right direction, now the episode names of the downloaded files will be correct, and the correct episodes will be skipped if the user decides to skip filler.

The episodes flag doesn't take ep_no into consideration, but I suppose that that's a separate issue

ArjixGamer:
woah, this is so nice, i think this should be a feature in all the providers, manually providing the episode number

AbdullahM0hamed:
`-e` flag now accounts for `ep_no`. Tested with: `-e 945:` (which would normally return an index error), `-e 945:946`, `-e 946`, and `-e :9`

PR #562, opened 2020-10-19
Add ability to queue anime 

### LINKED ISSUE #418: List of anime , auto shutdown

AbdullahM0hamed:
Closes #418.

The issue made 3 requests: 
• Auto-shutdown - this will not be added as per #419.
• An option to pause - this can be done by cancelling mid-download with aria2c, and re-running the command when you wish to continue.
• The ability to queue anime - this is what this PR is for.

Normal usage is unaffected. This PR adds a `-q` flag, which when used, allows the user to provide a comma delimited list of anime to queue: e.g: `anime dl -q "Naruto, Black Fox"`. When using the `-q`, if either the `-c` flag or the `-e` flag are used, they are also expected to be a comma delimited list of choices/episode selections and must match the number of queued anime.

I've tested and everything appears to work, but I recommend that others do rigorous testing as well.

### DIFF
--- a/anime_downloader/commands/dl.py
+++ b/anime_downloader/commands/dl.py
@@ -84,7 +84,7 @@ sitenames = [v[1] for v in ALL_ANIME_SITES]
 @click.option(
     "--queue", "-q",
     type=bool, is_flag=True,
-    help="Queue a series of anime"
+    help="Queue a list of anime"
 )
 @click.pass_context
 def command(ctx, anime_url, episode_range, url, player, skip_download, quality,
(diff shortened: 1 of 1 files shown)


vn-ki:
I don't like this tbh. What if there is a `,` in the title and the user wants to search that? The results would look confusing. 

I don't see how this is much different from 

```
anime dl 'naruto' -c 1 ; anime dl 'black fox' -c 1
```

either.

AbdullahM0hamed:
I have to agree lol. And queueing the commands instead of having a queue flag is way cleaner looking too

PR #568, opened 2020-11-05
Fix 9anime and animesuge

AbdullahM0hamed:
Thanks @theinternet (discord - no idea what his ghub name is) for the decoding code)


nate-moo:
@theinternetuser

anime_downloader/sites/animesuge.py
```diff
@@ -35,6 +34,81 @@ def _scrape_metadata(self):
 
 
 class AnimeSugeEpisode(AnimeEpisode, sitename='animesuge'):
+    """
+    Decodes the encoded string returned by /ajax/anime/episodes to the embed url.
+
+    Python implementation of decode9AnimeURL, it's basically identical to the js one
+    """ 
+    def decodeString(self, str):
```
vn-ki, line 42: If both are the same code, maybe move it util or something?

vn-ki (commented):
It's better to keep deobfuscation code close to it's source format, so it's okay I guess.

There isn't much point because they'll break this in a week with some update.

anime_downloader/sites/animesuge.py
AbdullahM0hamed: Made AnimeSugeEpisode inherit NineAnimeEpisode and got rid of the duplicate method

PR #571, opened 2020-11-12
Made the config command only allow one instance to run at a time.

ArjixGamer:


ArjixGamer:
This is an important change because if you run the config command in multiple instances there is a good chance it will corrupt the json.

anime_downloader/commands/config.py
```diff
@@ -1,6 +1,30 @@
...
+    for key, value in dict_pids.items():
+        if bool(value[1]):
+            if ('anime config' in ' '.join(value[1]) or
+                    'anime.exe config' in ' '.join(value[1])
+                        ) and ('python' in ' '.join(value[1])):
+                already_running = True
+    return already_running
+
+if check_running():
+    raise Exception('Another instance of "anime config" is already running!')
```
michaeldavidjohnson, line 27: "is active" or "is running" makes more sense imo

IguanasInPyjamas, line 27: "is active" or "is running" I think makes more sense here

anime_downloader/commands/config.py
```diff
@@ -1,6 +1,30 @@
...
+        for p in psutil.process_iter(attrs=["pid", "name", 'cmdline'])
+    }
+
+    if os.getpid() in dict_pids:
+        del dict_pids[os.getpid()]
+
+    for key, value in dict_pids.items():
+        if bool(value[1]):
+            if ('anime config' in ' '.join(value[1]) or
+                    'anime.exe config' in ' '.join(value[1])
```
IguanasInPyjamas, line 21: is there not a way just to regex out the words anime config, would remove the or statement.

anime_downloader/commands/config.py
ArjixGamer: indeed

anime_downloader/commands/config.py
ArjixGamer: this convo is already discussed in discord, just posting a screenshot here so that i can resolve it
image

ArjixGamer:
I think this PR is done.

ArjixGamer:
If there was smth for me to add that would be a docstring

ArjixGamer:
I think this PR is done.
If there was smth for me to add that would be two docstrings.

anime_downloader/util.py
```diff
@@ -459,3 +459,83 @@ def type_cast_value(self, ctx, value):
...
+                placeholder.append(
+                    Process(name=f[0], cmdline=f[1:-1], pid=int(f[-1])))
+            else:
+                placeholder.append(
+                    Process(name=f[0], cmdline=None, pid=int(f[-1])))
+    return placeholder
+
+
+def getAllProcesses_unix():
+    from sys import platform
```
IguanasInPyjamas, line 495: sys is also imported at this point, can't you just use sys.platform

anime_downloader/util.py
```diff
@@ -459,3 +459,83 @@ def type_cast_value(self, ctx, value):
...
+        line_list = line.split()
+        PID = line_list[1]
+        NAME = line_list[10:][0]
+        CMD = line_list[10:]
+        placeholder.append(Process(name=NAME, cmdline=CMD, pid=PID))
+    return placeholder
+
+
+def get_all_processes():
+    from sys import platform
```
IguanasInPyjamas, line 513: same point here

anime_downloader/util.py
```diff
@@ -459,3 +459,83 @@ def type_cast_value(self, ctx, value):
...
+        return getAllProcesses_unix()
+
+
+def is_running(regex, expected_matches):
+    """
+    Iterates through all the processes that are running
+    and returns a boolean if a process matches the regex passed
+    and the groups matched are equal to or more than the expected_matches.
+    """
+    import re
```
IguanasInPyjamas, line 526: re is already imported...

ArjixGamer:
done

ArjixGamer:
Because we were getting a lot of unreasonable errors for no apparent reason, I decided to just not support mac in this, if the user uses a mac they will not see any change to the behavior of anime dl from this PR. In other words this PR won't affect mac users in any kind of way.

anime_downloader/commands/config.py
```diff
@@ -1,7 +1,11 @@
-import click
+from anime_downloader.util import is_running
+if is_running(regex=r'python|anime|config', expected_matches=3):
+    raise Exception('Another instance of "anime config" is already running!')
```
AbdullahM0hamed, line 3: Move this code to the command method

anime_downloader/commands/ezdl.py
```diff
@@ -145,17 +151,18 @@ def command(ctx, anime_url, episode_range, player,
...
 
             try:
                 animes = util.parse_ep_str(anime, str(episode_range))
             except RuntimeError:
-                logger.error('No episode found with index {}'.format(episode_range))
+                logger.error(
+                    'No episode found with index {}'.format(episode_range))
                 continue
-            except:
+            except BaseException:
```
AbdullahM0hamed, line 165: This is pointless, did you aggressively autopep8?

anime_downloader/commands/watch.py
```diff
@@ -1,3 +1,8 @@
+from anime_downloader.util import is_running
+if is_running(regex=r'python|anime|watch', expected_matches=3):
+    raise Exception('Another instance of "anime watch" is already running!')
```
AbdullahM0hamed, line 3: Move this to the command method

anime_downloader/util.py
```diff
@@ -459,3 +459,84 @@ def type_cast_value(self, ctx, value):
...
+        })
+
+
+def getAllProcesses_Win32():
+    placeholder = list()
+    out = os.popen('WMIC path win32_process get Caption,Processid,Commandline').read(
+    ).split('\n')[::2][1:]
+    for line in out:
+        f = line.split()
+        if bool(f):
```
AbdullahM0hamed, line 484: Unnecessary bool cast - `if f:` should suffice. Also, consider naming the variable something meaningful

AbdullahM0hamed (changes requested):

anime_downloader/commands/ezdl.py
ArjixGamer: yes

PR #578, opened 2020-11-23
Throw UsageError if episode flag is used incorrectly

AbdullahM0hamed:


ArjixGamer:
are you sure this correctly throws a proper exception? (it does not throw the one you added in the PR at least)
image

ArjixGamer:
i mean, ok, it works if i do this
image
but you dont cover all the incorrect usages

PR #580, opened 2020-11-25
Improve test command functionality

czoins:


nate-moo:
When `anime test` finishes it prints a massive broken table so I may make a PR or suggest an edit to fix that, other than that It worked for me

anime_downloader/commands/test.py
```diff
@@ -3,66 +3,178 @@
...
+                logging.debug('Timeout during testing.')
+                click.secho(f + "Not working: Timeout. Use -t to specify longer waiting period.", fg="red")
+
+        except KeyboardInterrupt:
+            skip = click.confirm(f"\nSkip {thread.provider} and continue searching? (Press enter for Yes)", default=True)
+            if not skip:
+                break
+
+    if not no_results:
+        click.echo("\n" + util.format_matches(matches))
```
nate-moo, line 179: Maybe instead of trying to format the matches, just print `Anime Test Finished` because it will overtake the screen with a broken table without it. change it like this:
```click.echo("\n" + "Anime test finished")```
or if you still want it, make it a switch to enable it. Keep in mind the formatting of the table would have to be fixed.

**Comment by nate-moo** [comment]
Now that I think of it, the broken table may have something to do with threading, but I'm not too well versed in threading so take this with a grain of salt

**Comment by Skaytacium** [comment]
> Maybe instead of trying to format the matches, just print Anime Test Finished because it will overtake the screen with a broken table without it. change it like this:
click.echo("\n" + "Anime test finished")
or if you still want it, make it a switch to enable it.

Completely agree with this, a switch to **enable** table output instead of disabling it would be great because not everyone wants to check the results.

Another small modification could be trimming the number of results, for example, if `--trim 5` is specified, only show 5 results per site.

Great functionality though.

**Comment by nate-moo** [comment]
> Now that I think of it, the broken table may have something to do with threading, but I'm not too well versed in threading so take this with a grain of salt

I'm just an idiot, ignore this

**Comment by nate-moo** [comment]
@IguanasInPyjamas are you just gonna ignore this? -_-

**Comment by IguanasInPyjamas** [comment]
I haven't looked at it yet.

**Comment by IguanasInPyjamas** [comment]
Also there are no commits since your feedback, what do you actually want me to do that warrants tagging me?

**Review by AbdullahM0hamed — CHANGES_REQUESTED** [review]
@czoins `anime test -d` still outputs results unless it is utilised with the `-f` flag

**Comment by czoins** [comment]
The result tables are now disabled by default, use `-r` to enable them. It also performs a full search by default, use `-f` to have it prompt you to stop on anime match.
As for the formatting, the table isn't broken, it just doesn't nicely fit a small console with a large font, so consider having it on fullscreen at least if you want the results.

**Review by AbdullahM0hamed — APPROVED** [review]

File: anime_downloader/commands/test.py [hunk 2]
```diff
@@ -3,66 +3,181 @@
...
 
+    if providers:
+        providers = [p.strip() for p in providers.split(",")]
+        for p in providers:
+            if not p in sitenames:
+                raise click.BadParameter(f"{p}. Choose from {', '.join(sitenames)}")
+    else:
+        providers = sitenames
+        if not selenium:
+            providers.remove("kisscartoon")
```
IguanasInPyjamas, line 110: This needs to be removed, kisscartoon was already removed from the main branch.

czoins:
I removed the `--selenium` flag as well, since there are currently no providers using selenium.

IguanasInPyjamas:
I'm happy with this, tested cz's pr on his version on arch. Windows and 5.0.7 testing needs to be done before merge

AbdullahM0hamed (approved):

PR #588, opened 2020-12-10
Added arabic provider EgyAnime

ArjixGamer:


ArjixGamer:
i also added a new extractor, its for the website clipwatching

AbdullahM0hamed:
I think it'd be great if you handled cases like https://www.egyanime.com/watch?w=8085 by throwing a custom logging.error message to let the user know there is a copyright violation

AbdullahM0hamed:
For `anime dl -u naruto -p egyanime -c 5` - this gets https://www.egyanime.com/movie?do=10 instead of https://www.egyanime.com/movie?w=10 which then (probably) causes it to fail to extract the final link

IguanasInPyjamas (approved):
I'm fine with this, final say is down to @AbdullahM0hamed due to the m3u8s. 

Tested on Windows and Linux. could not find any issues.

PR #590, opened 2020-12-16
Change split anime test

AbdullahM0hamed:
Due to changes to the split_anime method - split_anime now uses specific attributes of the anime object to do the splitting (so episode selection can be more accurate), this means that passing a list object will lead to an AttributeError, therefore we should replace it with a mock object


IguanasInPyjamas:
ew mocks, but i'm fine with it

Issue #589, opened 2020-12-16
Launch options for mpv


akash-258:
Plz, tell me how can I launch MPV with a specific profile for "anime watch". Currently, it uses the default profile but I want it to launch with something like " mpv --profile=anime4k "

IguanasInPyjamas:
there's no mpv args being handled, can you not set default behaviour in mpv? I may look into adding this

akash-258:
I cannot add the 'anime4k' shader in the default config because it will affect other normal videos as well. So I set the shaders in input.conf but using that I have to activate the shader using that key in every episode I watch. So I was thinking of a way so that I can launch MPV with a --profile option, that way the shader will only work in anime watch.
Please implement this feature in some upcoming release as it will be really helpful or tell me some way it could be done in the current release.

akash-258:
And maybe the "anime4k" shader can be added to our wonderful application itself, which can be activated with a keypress.

theCow61:
There seems to be a "mpv-config.conf" in .config/anime-downloader folder though every time I edit it, my changes get overwrited but I think thats a good place to start for figuring this out and from the looks of it you can mess with key bindings in it.

0kalekale:
@IguanasInPyjamas
> there's no mpv args being handled, can you not set default behaviour in mpv? I may look into adding this

how about this:
define mpv args/flags somewhere in ~/.conifg/anime-downloader/config.json
```js
{
"mpv-args":
 [
    {"arg" : "--profile=anime4k"},
    {"arg" : "--slang=eng"}
 ]
}
```
and when you call ```anime watch <something>``` the args get passed into mpv from the config file

0kalekale:
@theCow61

> There seems to be a "mpv-config.conf" in .config/anime-downloader folder though every time I edit it, my changes get overwrited but I think thats a good place to start for figuring this out and from the looks of it you can mess with key bindings in it.

from [anime_downloader/players/mpv.py](https://github.com/anime-dl/anime-downloader/blob/master/anime_downloader/players/mpv.py):

```py
def get_mpv_configfile():
    # Read the user's input config file if it exists
    userconf = os.path.join(get_mpv_home(), 'input.conf')
    userconftext = ''
    if os.path.exists(userconf):
        with open(userconf, 'r') as userconfigfile:
            userconftext = userconfigfile.read()

    # Create a new config file to add anime-downloader specific key bindings
[...]

PR #593, opened 2020-12-22
Mpv launch arguments added.

theCow61:
Adds mpv launch arguments for issue: #589 . Run anime config, go to watch config options and mpv_arguments should now be an option. You can set this to --profile=Anime4k for example and mpv will launch with those arguments and any arguments you add ontop of it.


IguanasInPyjamas:
Seems fine. I'll do some testing

IguanasInPyjamas (approved):
Fine

PR #598, opened 2021-01-08
Fixed the Subsplease sources extraction

AditCh:
_scrape_episodes was saving incorrect keys, resulting in wrong fake urls
_get_sources used incorrect keys for magnet links


ArjixGamer:
Hmm, looks like subsplease changed their api response so thats why it broke, ill tag someone to merge this as it works

PR #614, opened 2021-01-29
Fix twist

IguanasInPyjamas:
This cdn can only get the backlog, not the airing cdn. They store airing on a different cdn it seems, It's kinda awkward to dynamically choose which cdn to switch, so I've just set the biggest library cdn

maybe we can make it as a config option or something?

### DIFF
--- a/anime_downloader/sites/twistmoe.py
+++ b/anime_downloader/sites/twistmoe.py
@@ -71,7 +71,7 @@ class TwistMoe(Anime, sitename='twist.moe'):
         episodes = episodes.json()
         logging.debug(episodes)
         self.title = anime_name
-        episode_urls = ['https://twistcdn.bunny.sh' +
+        episode_urls = ['https://cdn.twist.moe' +
                         decrypt(episode['source'].encode('utf-8')).decode('utf-8')
                         for episode in episodes]
 


AbdullahM0hamed:
As all the airing anime I've tried with this PR work as of now - I've opted to merge.

ArjixGamer:
or just scrape both at the same time...

IguanasInPyjamas:
> or just scrape both at the same time...

That seems inefficient at best and moronic at worst, the ideal situation would be to handle a 404 from the main cdn to then use the backup cdn. Obviously.

ArjixGamer:
> > or just scrape both at the same time...
> 
> That seems inefficient at best and moronic at worst, the ideal situation would be to handle a 404 from the main cdn to then use the backup cdn. Obviously.
> 
> Although it doesn't seem like it's too much of an issue, I have yet to find one that is on the other cdn and not the main one lol

Oh, it looks like I didn't quite understand your original message...
I thought you meant that any new anime is only available on the new cdn...

AbdullahM0hamed:
@ArjixGamer he did mean that - but how would you determine what's new? At the time of merging this was a non-issue because the one cdn worked for all airing anime anyway

ArjixGamer:
> @ArjixGamer he did mean that - but how would you determine what's new? At the time of merging this was a non-issue because the one cdn worked for all airing anime anyway
> 
> Edit: he meant that - it was supposed to be that, but for some reason it wasn't 

Just go to sleep...

AbdullahM0hamed:
Very soon I shall perform the wondrous time-skip magic known as "sleep"

PR #617, opened 2021-02-06
Adding support for android intents

nate-moo:


AbdullahM0hamed:
`anime dl naruto --play android -c 1 -p animefreak` despite being 1 episode long (movie), still waits for input

nate-moo:
Unless we do some checks to see how many episodes are in the show that may be a little difficult to fix as the current logic only looks for if an -e argument is applied and what it's contents are

anime_downloader/util.py
```diff
@@ -222,13 +222,19 @@ def print_episodeurl(episode):
...
             f'--title={title}', 
             f'--referrer={episode.source().referer}', 
             f'--user-agent={get_random_header()["user-agent"]}', 
             episode.source().stream_url])
+    elif player == "android":
+        if episodes == None or ':' in episodes:
+            p = subprocess.Popen(['am', 'start', '-a', 'android.intent.action.VIEW', '-t', 'video/*', '-d', f'{episode.source().stream_url}'])
+            input("Press enter to continue\n")
+        else:
+            p = subprocess.Popen(['am', 'start', '-a', 'android.intent.action.VIEW', '-t', 'video/*', '-d', f'{episode.source().stream_url}'])
```
AbdullahM0hamed, line 237: Don't repeat this line - do it before the if statement, and only put the input in the if statement

anime_downloader/commands/dl.py
```diff
@@ -130,7 +130,7 @@ def command(ctx, anime_url, episode_range, url, player, skip_download, quality,
             util.print_episodeurl(episode)
 
         if player:
-            util.play_episode(episode, player=player, title=f'{anime.title} - Episode {episode.ep_no}')
```
AbdullahM0hamed, line 133: Add something like `episode_ramge = f"0:{len(animes)}" if not episode_ramge else episode_range` before this line

AbdullahM0hamed (changes requested):

anime_downloader/commands/dl.py
nate-moo: Had to modify util.py to get this to fully work

anime_downloader/util.py
```diff
@@ -221,20 +221,24 @@ def print_episodeurl(episode):
...
+            f'--title={title}',
+            f'--referrer={episode.source().referer}',
+            f'--user-agent={get_random_header()["user-agent"]}',
             episode.source().stream_url])
+    elif player == "android":
+        p = subprocess.Popen(['am', 'start', '-a', 'android.intent.action.VIEW', '-t', 'video/*', '-d', f'{episode.source().stream_url}'])
+        if episodes == None or ':' in episodes:
+            if "0:1" in episodes:
+                pass
+            else:
```
AbdullahM0hamed, line 236: Unnecessary if-else, negate the condition by doing `if "0:1" not in episodes` instead

AbdullahM0hamed (changes requested):

anime_downloader/util.py
nate-moo: Didn't think about that, lemme go fix

AbdullahM0hamed:
Works, merging

PR #623, opened 2021-02-10
a small fix for shiro.is

ArjixGamer:
Smol fix :p

### DIFF
--- a/anime_downloader/sites/shiro.py
+++ b/anime_downloader/sites/shiro.py
@@ -1,5 +1,5 @@
-import re
 import logging
+import re
 from anime_downloader.sites.anime import Anime, AnimeEpisode, SearchResult
 from anime_downloader.sites import helpers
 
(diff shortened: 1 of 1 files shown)


ArjixGamer:
provider works fine even w/o this small fix
basically, if you copy an episode link, w/o this fix it would fail to find the anime
(yes, i know that pasting episode links will behave the same as pasting the anime link itself)

AbdullahM0hamed:
Tried with: `anime dl -u https://shiro.is/stream/dubu-xiaoyao-episode-72` got `KeyError: status` @ArjixGamer

ArjixGamer:
its bc the slug of the anime in that link is different from the slug of the anime in the api...
https://shiro.is/anime/dubbedu-xiaoyao

ArjixGamer:
nothing i can do, only way for me to get the slug is from the url...
scraping the website is not an option since its a dunamic only site

PR #629, opened 2021-02-13
removed unneeded regex from nyaa

ArjixGamer:


ArjixGamer:
css selectors are cooler than regex
prove me wrong

AbdullahM0hamed:
> css selectors are cooler than regex
prove me wrong

No, you are right.

PR #630, opened 2021-02-14
fixed animefrenzy

ArjixGamer:
note:
apparently it has exactly the same api as shiro.is but its just hosted somewhere else...
im really confused


ArjixGamer:
i think its safe to merge...
cause it literally has the same api as shiro...

ArjixGamer:
*patch-29 goes brrr*

AbdullahM0hamed:
Works

PR #633, opened 2021-02-15
improved the vidstreaming extractor (at least i think?)

ArjixGamer:
this should be tested to see if any providers break bc of it
(already broken providers dont count ;p)


ArjixGamer:
this PR should not exist....ill remake...its a mistake as is

ArjixGamer:
actually no, all i did was use a different link, so my PR was completely useless...

PR #634, opened 2021-02-15
Added greek provider An1me!

ArjixGamer:

### DIFF
--- a/anime_downloader/sites/an1menl.py
+++ b/anime_downloader/sites/an1menl.py
@@ -1,0 +1,49 @@
+import logging
+import re
+from anime_downloader.sites.anime import Anime, AnimeEpisode, SearchResult
+from anime_downloader.sites import helpers
+
+logger = logging.getLogger(__name__)
+
+
+class An1menl(Anime, sitename='an1me'):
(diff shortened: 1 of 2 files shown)


ArjixGamer:
purely regex

ArjixGamer:
since red did the soup, i made it ready for review

AbdullahM0hamed:
Don't we still have the issue of the links failing to download? 😔

ArjixGamer:
um, shh, shudup

AbdullahM0hamed:
> um, shh, shudup 

kako-6mmtatSBLB68SSYu

IguanasInPyjamas:
So this PR is open but fails to function?

AbdullahM0hamed:
It was initially a draft. I added a commit, and I think @ArjixGamer might've marked it as ready for review under the mistaken assumption that I'd fixed the problem in question

ArjixGamer:
Hmm, I'm not on pc so I won't test. But taking the links from ``-u`` and playing them with mpv worked, might be a referer issue? Like, it should have no referer instead of having itself as the referer?

PR #635, opened 2021-02-19
Add quality selection to animeout

AbdullahM0hamed:
Quality selection is desperately needed on this provider as animeout displays all episodes of all qualities in the same page and `_scrape_episodes` picks them all up as separate episodes. The difference can be seen with the command `anime dl -u hanayome -p animeout -c 1`


AbdullahM0hamed:
Since I prioritised `self.quality`, you should add `-q 720p` to the command you are using

lol480phuh:
It didn't work at first, so I had to reinstall using `pip3 install -U git+https://github.com/vn-ki/anime-downloader.git` and now it works, thank you!

PR #644, opened 2021-03-11
fixed 4anime & added an unpacker for packed js that uses pure python

ArjixGamer:
needs testing


ArjixGamer:
i did test it, but i can never be 100% sure so go wild testing

ArjixGamer:
too bad that more and more providers depend on nodejs...
maybe we should write our own unpacker for ``p.a.c.k.e.d.`` js

ArjixGamer:
found this beauty: https://github.com/beautify-web/js-beautify/blob/main/python/jsbeautifier/unpackers/packer.py
will do some tests on it

ArjixGamer:
aight, now 4anime doesnt use nodejs at all

ArjixGamer:
the unpacker can be used like this:
```python

if helpers.detect_packed_js(packed_js_string):
    result = helpers.unpack_packed_js(packed_js_string)
```

ArjixGamer:
changed it to import the unpackers instead of just copy pasting their code in anime dl

ArjixGamer:
this pr once again works

setup.py
```diff
@@ -31,6 +31,7 @@
         'requests-cache>=0.4.13',
         'tabulate>=0.8.3',
         'pycryptodome>=3.8.2',
+        'jsbeautifier==1.11.0'
```
AbdullahM0hamed, line 34: Doing this apparently has the potential to break this for some AUR and brew users. Perhaps move this to `extras_require` (add to existing, and one of its own)?

ArjixGamer: how tf does it have such a potential tho?
its not our problem if aur is fucked up, is it?

ArjixGamer: in the first place we dont the aur package is an unofficial one...

AbdullahM0hamed: It's unofficial? Strange that vn-ki would comment about it (see: https://github.com/anime-dl/anime-downloader/pull/443#issuecomment-662447368)

ArjixGamer: ok fine, you win, ill make it an optional dependency

ArjixGamer:
Ok, so I finished testing.
The unpacker returns its own input if jsbeautifier is not installed.
That way, it has the same result as it not being there at all.

PR #651, opened 2021-03-22
Adding video information flag

nate-moo:
please test on different OS's only tested on windows atm

### DIFF
--- a/anime_downloader/commands/dl.py
+++ b/anime_downloader/commands/dl.py
@@ -3,6 +3,7 @@ import os
 
 import click
 import requests_cache
+import requests
 
 from anime_downloader import session, util
 from anime_downloader.__version__ import __version__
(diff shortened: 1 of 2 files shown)


anime_downloader/commands/dl.py
```diff
@@ -121,6 +123,34 @@ def command(ctx, anime_url, episode_range, url, player, skip_download, quality,
     if speed_limit:
         logger.info("Speed is being limited to {}".format(speed_limit))
     for episode in animes:
+        if video_info:
+            import cv2
```
AbdullahM0hamed, line 127: Since the dependency is optional, perhaps try-except this import

anime_downloader/commands/dl.py
```diff
@@ -121,6 +123,34 @@ def command(ctx, anime_url, episode_range, url, player, skip_download, quality,
...
+            cv2video = cv2.VideoCapture(episode.source().stream_url)
+            try:
+                if cv2video.isOpened():
+                    try:
+                        framecount = cv2video.get(cv2.CAP_PROP_FRAME_COUNT ) 
+                        frames_per_sec = cv2video.get(cv2.CAP_PROP_FPS)
+                        duration = framecount / frames_per_sec
+                        logger.info(f"Video duration (sec): {duration}")
+                    except:
```
AbdullahM0hamed, line 136: Perhaps capture the expected exceptions `as e` and use if statements for the logic instead of multiple try-except blocks?

anime_downloader/commands/dl.py
```diff
@@ -79,10 +80,11 @@
     help="Set the speed limit (in KB/s or MB/s) for downloading when using aria2c",
     metavar='<int>K/M'
 )
+@click.option("--video-info", "-i", is_flag=True, help="enables getting video information")
```
AbdullahM0hamed, line 83: Capitalise the first letter to match the casing of other help messages. Additionally, add a sentence explain that this requires opencv-python

AbdullahM0hamed (changes requested):

PR #667, opened 2021-05-03
added uget

ArjixGamer:
Adds the external download manager uget.

Tested on windows with uget on path (portable version).


ArjixGamer:
should we wait for a linux test?

Issue #412, opened 2020-06-18
Animepahe server cannot be found


ayushjaipuriyar:
**Describe the bug**
Can't download from animepahe it just says

*ERROR: 503 Server Error: Service Temporarily Unavailable for url: https://animepahe.com/api?l=8&m=search&q=inuyasha*

**To reproduce**

anime dl inuyasha --provider animpahe -q 480p
and the error received is 

ERROR: 503 Server Error: Service Temporarily Unavailable for url: https://animepahe.com/api?l=8&m=search&q=inuyasha

this log level debug doesn't really work
<details>
<summary>LOG</summary>

```
On searching the url given in the error 

the browser shows 

{"total":1,"data":[{"id":1711,"slug":"inuyasha","title":"InuYasha (TV)","type":"TV","episodes":167,"status":"Finished Airing","season":"Fall","year":2000,"score":7.85,"poster":"https:\/\/i.animepahe.com\/posters\/9dc06bac320b0d778be645bff712f7d1.jpg","session":"d66bd27c-a074-9a98-14bf-0ad269def8fe","relevance":"38"}]}


```

Blatzar:
Yeah animepahe recently got cloudflare, will probably be able to fix with selenium.

Blatzar:
Looked into it, animepahe has cloudflare AND hcaptcha which will be promoted EVERY time you visit the site (wtf is this fort knox?). Cracking this is definitely possible, but it will take some time.

ayushjaipuriyar:
Okay I get the fact that it updates every time I open up the site but for some reason as such on the browser site works fine but it shows error on animedl

ayushjaipuriyar:
Also selenium doesnt fix the issue

Blatzar:
> Also selenium doesnt fix the issue

I'm well aware, it's just a step towards a fix. Before testing I thought it would solve since I've never seen a site use both cloud-flare and hcaptcha.

>AND hcaptcha

Selenium can get past the cloud-flare, @AbdullahM0hamed tool to bypass hcaptcha just needs to get added correctly and the site will probably work.

ayushjaipuriyar:
Ohh well let me try then 
Thanks for anime-downloader though

LordShenron:
Waiting for the fix now ^_^ 
Luckily I had downloaded a lot of anime for now.

ayushjaipuriyar:
I see no god up here .
having 2tb of anime and shows

[...]

PR #673, opened 2021-05-16
Fix animepahe

nate-moo:
fixes later parts of #412
animepahe returns a m3u8 so it isn't too helpful when downloading but ffmpeg can convert it to an mp4 with:

```
ffmpeg -headers "Referer: https://kwik.cx/" -i {link goes here} -c copy anime.mp4
```


nate-moo:
oops
this somehow got merged with another project

PR #678, opened 2021-05-19
fixed genoanime

ArjixGamer:


ArjixGamer:
oh yeah, forgot the import, thx red

PR #679, opened 2021-05-20
Reintroduce support for 9anime sites

### LINKED ISSUE #599: animedl can't download anything from 9anime,
**Describe the bug**

i try to download an anime using this repo, but it failed, 
Screenshot 2021-01-09 at 22 28 02

**To reproduce**

```

```

<details>
<summary>LOG</summary>

```



```

ekrekeler:
Closes #599

For now, same as before, only tested using Streamtape server.

Please suggest changes as needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ Yeah. Me too! That's why this tool exists.
 **Details about the sites can be found in [FAQ](https://github.com/vn-ki/anime-downloader/wiki/FAQ)**
 
 
-- 4Anime
+- AnimePahe
 - AnimTime
 - AnimeBinge
 - Animedaisuki


ArjixGamer:
wow, I never thought that someone would actually use selescrape

ArjixGamer:
I tested and it does indeed return a streamtape link
its just that streamtape doesnt work as an extractor

ekrekeler:
So the streamtape extractor doesn't work sometimes because the server doesn't like the ancient user agent presented by the get helper function. I overrided the user agent in streamtape.py to a more recent one, and haven't seen the 503 error since.

Do you think it makes sense to define a user agent in streamtape.py, or should this be a global setting? Why is the user agent so old anyways? Using python 3.9.5 on WSL2, this is the user agent shown in debug:
`Mozilla/5.0 (Windows NT 6.3; WOW64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/41.0.2225.0 Safari/537.36`

ArjixGamer:
woah

ArjixGamer:
wait no, im dumb, you werent talking about that

ekrekeler:
> Yeah, it was because of selescrape

I'm not sure it is, because there is no `sel=True` in extractors/streamtape.py. So it shouldn't be using selescrape for this request.

> wait no, im dumb, you werent talking about that

Yeah I just want to know if I should be defining a different header in streamtape.py. I've seen instances where sometimes it works when there are two user-agents defined in headers. So this is a separate issue. There are certain user-agents in that random list that the streamtape server doesn't like. So I can either:

1. Find the user agents that don't work and either update or remove them from the list
2. Select a user agent that works and use that for every request in streamtape.py

ArjixGamer:
> Why I am asking is I don't know is if that list of user agents has a reason for being so out-of-date compared to current user agents, and if changing that list will break (or fix) other sites.

You can go ahead and update that list if you want to.
As far as I know that list has not been updated for a long time.

ekrekeler:
Okay I've updated all the user agents I could find in the code. Tested using `anime test` and I saw no differences before and after modifying the user agents.

I also tweaked the decodeString method in nineanime.py to remove some extra characters from the URL that shouldn't be there. I have no idea how the encoding for the 9anime API works so I'm just using character matching after the string is decoded instead.

Couple of things to note for next steps:

- There is still a duplicate user-agent header when using sites/helpers/request.py. While this doesn't pose an issue for 9anime specifically, it should probably be addressed at some point in a separate issue.
- It would be wise to use some sort of external library or API to provide an up-to-date list of user agents so that you don't have to update the list manually. I looked briefly at some solutions but wasn't able to find a suitable one. I think the most reasonable is the WhatIsMyBrowser.com API, however this would require users to sign up for an API key before using this software.

ekrekeler:
There is someone else working on an alternate pull request for 9anime, #682 .

It seems that selenium may not be needed to get 9anime working if cloudflare is bypassed. I did not realize this. I will keep this open for the time being, it seems there are still some things that need to be worked out for that pull request. But avoiding the requirement of selenium is preferable.

ArjixWasTaken:
So uhh...
According to @justfoolingaround 9anime is changing its protection on a daily basis.
So if this PR is ready I'll merge.

ArjixWasTaken:
Oh, it doesn't work.
They have a "verified" url parameter now...

PR #681, opened 2021-05-23
Added provider wcostream

ArjixWasTaken:


ArjixWasTaken:
welp, I added it to config as well,
the mcloud server always has just one file
the vidstreampro server always has 2 files, one mp4 and one m3u8

(both servers have the same api, so I guess its just a backup server)

ArjixWasTaken:
Oh yeah, even tho one of the servers is named "VidStream" it's not the vidstream we all know, it just happens to have the same name.

PR #689, opened 2021-06-21
Improved the logic of the extractor selection in animerush

ArjixWasTaken:
and also added a proper error message in the extractor yourupload to say file not found.


ArjixWasTaken:
hmm, wait a min, ill change how yourupload handles the error

ArjixWasTaken:
Ok, it now acts the same way as the other extractors

PR #693, opened 2021-06-24
Add AnimeStar to sites

displicit:
Its API was begging to be used so much, that I found this repository.
Note that per JSON it scrapes resources like gogoanime; `/get-stream-links` takes its URL as a parameter.
Quality on every file may vary and there is no indication how it does.


ArjixWasTaken:
Gets approved by me just because you added metadata.
(my approval means nothing)

PR #701, opened 2021-07-09
Fix 'CachedResponse has no len()' error

### LINKED ISSUE #694: ERROR
Hello once again I got this error message 
`
ERROR: object of type 'CachedResponse' has no len()
`
here the full terminal output
Screenshot from 2021-06-26 20-56-01

Sadly I dont know if there is any Debug function or any logs.
BR.

RPMYT:
Closes #694.

### DIFF
--- a/anime_downloader/sites/helpers/request.py
+++ b/anime_downloader/sites/helpers/request.py
@@ -146,7 +146,10 @@ def soupify(res):
     -------
     BeautifulSoup.Soup
     """
-    soup = BeautifulSoup(res.text, 'html.parser')
+    if isinstance(res, str):
+        soup = BeautifulSoup(res, 'html.parser')
+    else:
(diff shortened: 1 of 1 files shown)


ArjixWasTaken:
I cba to test this (bc I'm away from home).
But the most major flaw that I found in this PR is that it doesn't deal with the issue at hand.


This error occurs because of the package we use to cache the requests we make, `requests-cache`.

 That package has moved a class from one place to another.
So anime-dl fails to cache any request.

To fix this we can either force the use of an older version of `requests-cache` or to add a try except in order to keep compatibility with both the old and new versions of `requests-cache`.


So what you should be looking to fix is in `session.py`. 

~~There is a small chance that I'm wrong about this, but installing an older version of that package fixes this error.~~

PR #703, opened 2021-07-21
Fixed tenshi.moe

ArjixWasTaken:
+ fixed video extraction
+ also paginates the search page now


ArjixWasTaken:
I did test on my end, but any more further tests are welcome.

PR #708, opened 2021-08-06
Arjix was taken patch 1

ArjixWasTaken:
Removed 4anime.


AbdullahM0hamed:
4anime, more like gone *4ever*

ArjixWasTaken:
> 4anime, more like gone _4ever_

LMAO

PR #709, opened 2021-08-06
Check for new versions

ArjixWasTaken:
This simply compares the installed version with the latest version hosted on github.
(I plan to add Pypi later as well.)


ArjixWasTaken:
It would be nice to have a way to detect if this package is installed via pip or if it was cloned and installed.

PR #725, opened 2021-12-14
I've edited the external_download function

Alsira:
Hello, I've modified a bit of the external downloader code so it can handle some simple errors so far. It works as normal and everything. It should recover on linux if aria2c needs root access to do some I/O and even create directories now.

Thanks for this awesome repo though. 😁😁

### DIFF
--- a/anime_downloader/util.py
+++ b/anime_downloader/util.py
@@ -393,18 +393,48 @@ def external_download(cmd, episode, file_format, speed_limit, path=''):
     cmd = format_command(cmd, episode, file_format, speed_limit, path=path)
 
     logger.debug('formatted cmd: ' + ' '.join(cmd))
-
+    
     if cmd[0] == 'open':  # for torrents
         open_magnet(cmd[1])
+    
     else:
-        p = subprocess.Popen(cmd)
-        return_code = p.wait()
-
-        if return_code != 0:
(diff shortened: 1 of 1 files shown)


nate-moo:
I don't believe that running shit as root is in the best interest of the end user as that could potentially brick or cause a malfunction in the system, especially if the user isn't particularly adept at computers. 

For example aria could overwrite an essential file if the user is trying to download to a restricted directory.

ArjixWasTaken:
> I don't believe that running shit as root is in the best interest of the end user as that could potentially brick or cause a malfunction in the system, especially if the user isn't particularly adept at computers.
> 
> For example aria could overwrite an essential file if the user is trying to download to a restricted directory.

You think a linux user wouldn't be adept at computers?

nate-moo:
> > I don't believe that running shit as root is in the best interest of the end user as that could potentially brick or cause a malfunction in the system, especially if the user isn't particularly adept at computers.
> > For example aria could overwrite an essential file if the user is trying to download to a restricted directory.
> 
> You think a linux user wouldn't be adept at computers?

yes, some might. Don't underestimate the power of the human brain, and plus installing linux is much easier now as it can basically just be 8 1/2 clicks and you're installed

